PR #75, opened 2019-05-07
Updated call GetGroupByName from AD to user API. Updated tests.

RocheL:
VIH-3839: Initial Version - Update the call for GetGroupByNamefrom AD to user API.
### Change description ###
This PR has the api call change for 'GetGroupByName'
```
[ ] Yes
[x] No
```


AdminWebsite/AdminWebsite.IntegrationTests/Controllers/ControllerTestsBase.cs
```diff
@@ -20,19 +23,27 @@ public class ControllerTestsBase
...
                 WebHost.CreateDefaultBuilder()
                     .UseEnvironment("Development")
                     .UseKestrel(c => c.AddServerHeader = false)
-                    .UseStartup<Startup>();
+                    .UseStartup<Startup>().ConfigureServices(AddMocks);
             _server = new TestServer(webHostBuilder);
             GetClientAccessTokenForBookHearingApi();
         }
 
+        private void AddMocks(IServiceCollection obj)
```
ddikman: As mentioned, if this doesn't work anyway, let's revert it and do the test as unit test only.

RocheL: moved to unit test

PR #78, opened 2019-05-15
VIH-3839 :Feature/call user api from admin web

rehmanab:
**Before creating a pull request make sure that:**

- [x] commit messages are meaningful and follow good commit message guidelines
- [ ] README and other documentation has been updated / added (if needed)
- [ x] tests have been updated / new tests has been added (if needed)

Please remove this line and everything above and fill the following sections:

### Change description ###

```
[ ] Yes
[ ] No
```


AdminWebsite/AdminWebsite/Helper/AdministratorRoleClaimsHelper.cs
```diff
@@ -0,0 +1,65 @@
...
+            UserCaseTypes = userCaseTypes?.Value?.Split(",".ToCharArray(), StringSplitOptions.RemoveEmptyEntries) ??
+                            Enumerable.Empty<string>();
+        }
+
+        public bool IsVhOfficerAdministratorRole { get; set; }
+        public bool IsCaseAdministratorRole { get; set; }
+        public bool IsAdministratorRole { get; set; }
+        public IEnumerable<string> UserCaseTypes { get; set; }
+
+        public IEnumerable<Claim> GetAdministratorClaims()
```
ddikman: Since all claims would be for administrators, maybe we don't need to have the "Administrator" in the name here

AdminWebsite/AdminWebsite/Helper/AdministratorRoleClaimsHelper.cs
```diff
@@ -0,0 +1,65 @@
+﻿using AdminWebsite.Security;
+using AdminWebsite.Services.Models;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Claims;
+
+namespace AdminWebsite.Helper
+{
+    public class AdministratorRoleClaimsHelper
```
ddikman: Do we need the `Helper` in the name or does the name excluding `Helper` explain properly what it does?

AdminWebsite/AdminWebsite.UnitTests/Helper/AdministratorRoleClaimsHelperTest.cs
```diff
@@ -0,0 +1,91 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Security.Claims;
+
+namespace AdminWebsite.UnitTests.Helper
+{
+    public class AdministratorRoleClaimsHelperTest
+    {
+        [Test]
+        public void GetAdministratorClaims_Returns_Claims()
```
ddikman: I know you asked about this before and I don't think I gave a proper answer, since we're following the `should_something_when/given_something` pattern can we keep that format on all the tests?

AdminWebsite/AdminWebsite.UnitTests/Helper/AdministratorRoleClaimsHelperTest.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            var userGroupData = new UserGroupData
+            {
+                UserRole = "Role1",
+                CaseTypes = new List<string> { "Case1", "Case2" }
+            };
+
+            var claims = new AdministratorRoleClaimsHelper(userGroupData).GetAdministratorClaims().ToList();
+
+            claims.Should().NotBeNull();
```
ddikman: More a preference thing but, to reduce the code, I generally find that if you're testing that it contains an element further down or using the `claims` variable, you'll in a sense get the assertion "for free". I mean, the test will fail with a stacktrace of exactly what was null if it isn't set. Same with the count `HaveCountGreaterThan(0)`, that'll be implicitly checked when you're pulling out a `FirstOrDefault`

AdminWebsite/AdminWebsite.UnitTests/Helper/AdministratorRoleClaimsHelperTest.cs
```diff
@@ -0,0 +1,91 @@
...
+            claims.Should().NotContainNulls();
+            claims.Should().HaveCountGreaterThan(0);
+            var userCaseTypesClaim = claims.FirstOrDefault(x => x.Type == "UserCaseTypes");
+            userCaseTypesClaim.Should().NotBeNull();
+            userCaseTypesClaim?.Value.Should().Be("Case1,Case2");
+        }
+
+        [TestCase("VhOfficer", true)]
+        [TestCase("SomeRole", false)]
+        public void IsVhOfficerAdministratorRole_Properties_Are_Correctly_Set(string role, bool expectedClaimValue)
```
ddikman: a bit hard to read this, both in terms of what the test is supposed to test and what it actually does. it looks a bit like you're testing two things

AdminWebsite/AdminWebsite.UnitTests/Security/UserIdentityTests.cs
```diff
@@ -1,43 +1,90 @@
...
+            var userIdentity = new UserIdentity(user);
+
+            userIdentity.Should().NotBeNull();
+            userIdentity.IsAdministratorRole().Should().Be(state ?? false);
+        }
+
+        [TestCase(true)]
+        [TestCase(false)]
+        [TestCase(null)]
+        public void IsVhOfficerAdministratorRole_Returns_Correct_Value(bool? state)
```
ddikman: What's `Correct_Value`?

AdminWebsite/AdminWebsite/Security/ClaimsMemoryCacheProvider.cs
```diff
@@ -0,0 +1,28 @@
...
+namespace AdminWebsite.Security
+{
+    public interface IClaimsCacheProvider
+    {
+        Task<IEnumerable<Claim>> GetOrAdd(string key, Func<string, Task<IEnumerable<Claim>>> valueFactory);
+    }
+
+    public class MemoryClaimsCacheProvider : IClaimsCacheProvider
+    {
+        private readonly ConcurrentDictionary<string, Task<IEnumerable<Claim>>> _cache;
```
ddikman: Feels jiffy to cache tasks. A task is a function right, which feels like you can't cache that, you can only cache results of something no?

AdminWebsite/AdminWebsite/Security/UserIdentity.cs
```diff
@@ -1,78 +1,56 @@
...
-            var groupClaims = _currentUser.Claims.Where(x => x.Type == "groups").ToList();
-            return groupClaims.Select(x => _userAccountService.GetGroupById(x.Value).DisplayName).ToList();            
+            return _administratorRoleClaimsHelper.UserCaseTypes;
         }
 
         /// <inheritdoc />
         public IEnumerable<string> GetAdministratorCaseTypes()
         {
-            return GetGroupDisplayNames().Where(group => AcceptedAdministratorRoles.Contains(group));
+            return _administratorRoleClaimsHelper.UserCaseTypes.Where(group => AcceptedAdministratorRoles.Contains(group));
```
ddikman: Using the `UserCaseTypes` you don't need to compare it with a list of roles anymore, the list will be the types.

AdminWebsite/AdminWebsite/Security/UserRole.cs
```diff
@@ -0,0 +1,14 @@
...
+namespace AdminWebsite.Security
+{
+    public enum UserRole
+    {
+        [Display(Name = "None")] None = 0,
+        [Display(Name = "VH Officer")] VhOfficer = 1,
+        [Display(Name = "Representative")] Representative = 2,
+        [Display(Name = "Individual")] Individual = 3,
+        [Display(Name = "Judge")] Judge = 4,
+        [Display(Name = "Case Admin")] CaseAdmin = 5
```
ddikman: This is just me not knowing, what do we use the display name for?

AdminWebsite/AdminWebsite/Startup.cs
```diff
@@ -123,5 +131,20 @@ public void Configure(IApplicationBuilder app, IHostingEnvironment env)
 
             app.UseMiddleware<ExceptionMiddleware>();
         }
+
+        private static Func<TokenValidatedContext, Task> OnTokenValidated()
+        {
+            return async ctx =>
```
ddikman: Instead of returning a function here, just rewrite the function to be the delegate?

AdminWebsite/AdminWebsite/Startup.cs
```diff
@@ -68,12 +71,17 @@ private void RegisterAuth(IServiceCollection serviceCollection)
...
                 options.DefaultAuthenticateScheme = JwtBearerDefaults.AuthenticationScheme;
                 options.DefaultChallengeScheme = JwtBearerDefaults.AuthenticationScheme;
+
             }).AddJwtBearer(options =>
             {
                 options.Authority = securitySettings.Authority;
                 options.TokenValidationParameters.ValidateLifetime = true;
                 options.Audience = securitySettings.ClientId;
                 options.TokenValidationParameters.ClockSkew = TimeSpan.Zero;
+                options.Events = new JwtBearerEvents
```
ddikman: You could even create your own class here, subclassing JwtBearerEvents and including a method for the OnTokenValidated handler. That way you can discreetly tuck away the functionality somewhere not in Startup.cs.

AdminWebsite/AdminWebsite/Security/CachedUserClaimBuilder.cs
```diff
@@ -0,0 +1,34 @@
...
+        public CachedUserClaimBuilder(IClaimsCacheProvider claimsCacheProvider, IUserAccountService userAccountService)
+        {
+            _claimsCacheProvider = claimsCacheProvider;
+            _userAccountService = userAccountService;
+        }
+
+        public async Task<IEnumerable<Claim>> BuildAsync(string username, string cacheKey)
+        {
+            return await _claimsCacheProvider
+            .GetOrAdd(cacheKey, async key => new AdministratorRoleClaimsHelper
```
ddikman: Feels like this code should be somewhere outside the cache. This is the functionality of how you actually get the user profile, not just how you cache it?

AdminWebsite/AdminWebsite/Helper/AdministratorRoleClaimsHelper.cs
```diff
@@ -0,0 +1,65 @@
...
+        public IEnumerable<string> UserCaseTypes { get; set; }
+
+        public IEnumerable<Claim> GetAdministratorClaims()
+        {
+            return _administratorClaims;
+        }
+
+        private static bool GetClaimBoolValue(Claim claim)
+        {
+            return !string.IsNullOrWhiteSpace(claim?.Value) &&
```
ddikman: Should we not raise an exception if we're sending in a null claim here? Will that even ever happen, since this is a private helper?

AdminWebsite/AdminWebsite/Helper/AdministratorRoleClaimsHelper.cs
```diff
@@ -0,0 +1,65 @@
...
+            var claimsList = claims.ToList();
+            var vhAdminRoleClaim = claimsList.FirstOrDefault(x => x.Type == VhOfficerAdministratorClaimName);
+            var caseAdminRoleClaim = claimsList.FirstOrDefault(x => x.Type == CaseAdministratorClaimName);
+            var adminRoleClaim = claimsList.FirstOrDefault(x => x.Type == AdministratorClaimName);
+            var userCaseTypes = claimsList.FirstOrDefault(x => x.Type == UserCaseTypesClaimName);
+
+            IsVhOfficerAdministratorRole = GetClaimBoolValue(vhAdminRoleClaim);
+            IsCaseAdministratorRole = GetClaimBoolValue(caseAdminRoleClaim);
+            IsAdministratorRole = GetClaimBoolValue(adminRoleClaim);
+            UserCaseTypes = userCaseTypes?.Value?.Split(",".ToCharArray(), StringSplitOptions.RemoveEmptyEntries) ??
```
ddikman: This is an interesting example from a coverage type of perspective :)

It looks like this is joining together quite a few test cases:
should_have_no_case_types_if_missing_case_types_claim
should_have_no_case_types_if_case_types_claim_is_null
should_have_a_list_of_case_types_if_claims_contains_comma_seperated_list_of_case_types

We could have a chat of these cases though, I'm thinking we could simplify it by making a few assumptions

AdminWebsite/AdminWebsite/Helper/AdministratorRoleClaimsHelper.cs
rehmanab: not sure, without helper dont know what it does...any suggestions ?

AdminWebsite/AdminWebsite.UnitTests/Helper/AdministratorRoleClaimsHelperTest.cs
rehmanab: done

AdminWebsite/AdminWebsite.UnitTests/Helper/AdministratorRoleClaimsHelperTest.cs
rehmanab: True but I normally like to know on what assertion failed quickly as opposed to reading the stack trace from an exception that have been thrown... It's sort of like Fail Fast ....maybe..

AdminWebsite/AdminWebsite.UnitTests/Helper/AdministratorRoleClaimsHelperTest.cs
rehmanab: Maybe, but you have to read it as its running 2 unit tests, 1. role = VhOfficer, 2 = SomeRole.
Reduces duplication. its like specflow examples, running the test multiple time for different criteria.

AdminWebsite/AdminWebsite.UnitTests/Security/UserIdentityTests.cs
rehmanab: done

AdminWebsite/AdminWebsite/Security/ClaimsMemoryCacheProvider.cs
rehmanab: No, Task is just an object that defines the work that needs to be done. When you get the task back from the cache, it will be executed, whether awaited or not.  And here we await it.

AdminWebsite/AdminWebsite/Security/UserIdentity.cs
rehmanab: Good catch!!

AdminWebsite/AdminWebsite/Security/UserRole.cs
rehmanab: Don't use it.. removed!

AdminWebsite/AdminWebsite/Startup.cs
rehmanab: Done

AdminWebsite/AdminWebsite/Startup.cs
rehmanab: Yeh I did think about that but no point at the moment as its only ever used here and not reusable ever probably. Can push it out if the need arises ?

AdminWebsite/AdminWebsite/Security/CachedUserClaimBuilder.cs
rehmanab: Yeh, lets do it.. But where ?

AdminWebsite/AdminWebsite/Helper/AdministratorRoleClaimsHelper.cs
rehmanab: If a Null claim come in, this method will return false, I think that's what we want as if a claim from the token/identity doesn't exist for any reason therefore we should return false.. ? Not Throw exception... ?

AdminWebsite/AdminWebsite/Helper/AdministratorRoleClaimsHelper.cs
rehmanab: Good spot, I have created  unit test
1) should_have_no_case_types_if_missing_case_types_claim

should_have_a_list_of_case_types_if_claims_contains_comma_seperated_list_of_case_types is already covered.

should_have_no_case_types_if_case_types_claim_is_null -> value of claim can not be null.

AdminWebsite/AdminWebsite/Helper/AdministratorRoleClaimsHelper.cs
rehmanab: Changed to Claims {get;}

AdminWebsite/AdminWebsite/Security/ClaimsMemoryCacheProvider.cs
rehmanab: I have changed this to cache the Task<UserRole>, and then parse claims afterwards.

AdminWebsite/AdminWebsite/Helper/AdministratorRoleClaimsHelper.cs
rehmanab: Changed this to throw ArgumentNullException when a admin claim is not present, forcing us to use the this class

AdminWebsite/AdminWebsite/Helper/AdministratorRoleClaimsHelper.cs
rehmanab: I have changed most of the tests to be more like testing the behavior, please check them out

AdminWebsite/AdminWebsite.IntegrationTests/Helper/MockBackchannel.cs
```diff
@@ -0,0 +1,25 @@
...
+            {
+                return await EmbeddedResourceReader.GetOpenIdConfigurationAsResponseMessage("microsoft-openid-config.json");
+            }
+
+            if (request.RequestUri.AbsoluteUri.Equals("https://inmemory.microsoft.com/common/discovery/keys"))
+            {
+                return await EmbeddedResourceReader.GetOpenIdConfigurationAsResponseMessage("microsoft-wellknown-keys.json");
+            }
+
+            throw new NotImplementedException();
```
ddikman: Maybe `InvalidOperationException` or a message included explaining what's not implemented? If we get in here for some reason it's probably useful to let the caller know why they've done something wrong.

AdminWebsite/AdminWebsite.UnitTests/Helper/AdministratorRoleClaimsTest.cs
```diff
@@ -0,0 +1,117 @@
...
+            parsed.IsCaseAdministratorRole.Should().Be(expectedValue);
+        }
+
+        [TestCase(UserRoleType.None, false)]
+        [TestCase(UserRoleType.VhOfficer, true)]
+        [TestCase(UserRoleType.CaseAdmin, true)]
+        [TestCase(UserRoleType.Individual, false)]
+        [TestCase(UserRoleType.Judge, false)]
+        [TestCase(UserRoleType.Representative, false)]
+        public void should_set_isadministratorrole_property(UserRoleType role, bool expectedValue)
```
ddikman: Mm, this makes a lot more sense to me (but then again, that might be because we've spent hours talking about it :P )

PR #96, opened 2019-06-18
Vih 4460 service

ddikman:

### Change description ###
Added in a testing layer to abstract the actual service from the cache.

```
[ ] Yes
[X] No
```


AdminWebsite/AdminWebsite/ClientApp/src/app/suitability/services/questionnaire-api.service.ts
```diff
@@ -0,0 +1,48 @@
...
+        return page;
+    }
+
+    private map(response: ParticipantSuitabilityAnswerResponse): ParticipantQuestionnaire {
+        return new ParticipantQuestionnaire({
+            displayName: `${response.first_name} ${response.last_name}`,
+            caseNumber: response.case_number,
+            hearingRole: response.hearing_role,
+            representee: response.representee,
+            hearingId: 'missing',
```
ddikman: Should we delete this if it's not part of the response from the BFF?

valentinabennett: Yes, we don't use it

PR #100, opened 2019-06-20
mapping suitability answers models

valentinabennett:
WIP  
### Change description ###
Mapping api client model to Suitability answer and questions model. 

```
[ ] Yes
[ X] No
```


AdminWebsite/AdminWebsite/ClientApp/src/app/suitability/answer-list-entry/answer-list-entry.component.ts
```diff
@@ -22,6 +22,10 @@ export class AnswerListEntryComponent implements OnInit {
     this.expanded = !this.expanded;
   }
 
+  get isRepresentative(): boolean {
+    return this.questionnaire.hearingRole.toLowerCase().indexOf('solicitor') >= 0;
```
rehmanab: hearing role may be undefined ?

ddikman: it is a mandatory field from the api

ddikman: however, can we avoid checking for solicitor here? it means that once we introduce another non-individual type, this will need to change.

I wonder, could we check if the `representee` field is set instead? so if you have any data there, we display it as `representing` otherwise as non-representing. What do you think?

AdminWebsite/AdminWebsite/ClientApp/src/app/suitability/services/questionnaire-api.service.ts
```diff
@@ -3,45 +3,121 @@ import { ParticipantSuitabilityAnswerResponse, SuitabilityAnswerResponse } from
...
-    }
-
-    private mapAnswer(answer: SuitabilityAnswerResponse): SuitabilityAnswer {
-        return new SuitabilityAnswer({
-            question: `Question text for ${answer.key}`,
-            answer: answer.answer,
-            notes: answer.extended_answer
+  constructor(private client: BHClient) { }
+
+  individualSuitabilityKeysOrder = [
```
ddikman: can you break out all of these to a separate service/class?

AdminWebsite/AdminWebsite/ClientApp/src/app/suitability/services/questionnaire-api.service.ts
```diff
@@ -3,45 +3,121 @@ import { ParticipantSuitabilityAnswerResponse, SuitabilityAnswerResponse } from
...
+  private map(response: ParticipantSuitabilityAnswerResponse): ParticipantQuestionnaire {
+    return new ParticipantQuestionnaire({
+      displayName: `${response.first_name} ${response.last_name}`,
+      caseNumber: response.case_number,
+      hearingRole: response.hearing_role,
+      representee: response.representee,
+      participantId: response.participant_id,
+      updatedAt: response.updated_at,
+      answers: this.mapAnswerGroups(response.answers,
+        response.hearing_role && response.hearing_role.toLowerCase()
```
ddikman: So here we can see again how the hearing role needs to be checked. Can we put this check into a service at least?

AdminWebsite/AdminWebsite/ClientApp/src/app/suitability/services/questionnaire-api.service.ts
```diff
@@ -3,45 +3,121 @@ import { ParticipantSuitabilityAnswerResponse, SuitabilityAnswerResponse } from
...
+          .indexOf('solicitor') > -1),
+    });
+  }
+
+  private mapAnswerGroups(answers: SuitabilityAnswerResponse[], isRepresentative: boolean): SuitabilityAnswerGroup[] {
```
ddikman: instead of passing in a representative, what do you think of having a mapper class to pass in?

```typescript
abstract class QuestionsMapper {
   map(questionId: string): string;
}

class RepresentativeMapper extends QuestionsMapper { ... }
```

This would make adding in translations or new common mapping very easy without having to modify the service.

Then you would have something like:
```typescript
this.mapAnswerGroups(response.answers, isRepresentative(response) ? representativeMapper : individualMapper)
```

or better yet:
```typescript
this.mapAnswersGroup(response.answers, this.getMapper(response))
```

AdminWebsite/AdminWebsite/ClientApp/src/app/suitability/services/questionnaire-api.service.ts
```diff
@@ -3,45 +3,121 @@ import { ParticipantSuitabilityAnswerResponse, SuitabilityAnswerResponse } from
...
+    return [
+      new SuitabilityAnswerGroup({
+        title: 'About you and your equipment',
+        answers: !isRepresentative ? this.mapAnswers(answers, this.individualSuitabilityKeysOrder, this.questionsIndividual)
```
ddikman: if using a mapper I think you may get away not having to pass in these order lists as well

AdminWebsite/AdminWebsite/ClientApp/src/app/suitability/services/suitability-answer.mapper.ts
```diff
@@ -0,0 +1,25 @@
...
+  AboutYourClient: 'ABOUT_YOUR_CLIENT',
+  ClientAttendance: 'CLIENT_ATTENDANCE',
+  HearingSuitability: 'HEARING_SUITABILITY',
+  Room: 'ROOM',
+  Camera: 'CAMERA_MICROPHONE',
+  Computer: 'COMPUTER'
+};
+
+export const SelfTestQuestionKeys = {
+  SeeYourself: 'SEE_YOURSELF',
```
ddikman: Looks good, I wonder if we should prefix them as well so we know they're part of the self-test

The runway team usually talks about the self-test in terms of "kit check", so maybe:
```
export const SelfTestQuestionKeys = {
SeeYourself: 'KIT_SEE_YOURSELF',
SeeYourself: 'KIT_MICROPHONE', // I think we're testing the mic here rather than the speakers?
SeeYourself: 'KIT_SEE_HEAR_CLEARLY',
}
```

AdminWebsite/AdminWebsite/ClientApp/src/app/suitability/answer-list-entry/answer-list-entry.component.ts
valentinabennett: it will be exception, should not be, but will check anyway

AdminWebsite/AdminWebsite/ClientApp/src/app/suitability/services/questionnaire-api.service.ts
valentinabennett: created class

valentinabennett: done

PR #102, opened 2019-06-21
Added deduplication to service

ddikman:

### Change description ###
Added in deduplication to questionnaire service.

This tracks all the questionnaires we've returned in the service and makes sure we don't return the same questionnaire twice.

Duplication can happen with cursor-based endpoints because of changes to the items that affect their ordering. Currently, the questionnaires are ordered by the latest added and no questionnaires will be "added in the past" so we would never see this behaviour with the current ordering because any changes would appear "before" our last cursor. But in the case of ordering being put in this is required.

```
[ ] Yes
[X] No
```


ddikman:
We had a quick talk and agreed we can skip this for now as the duplication won't occur.

PR #111, opened 2019-07-01
Refactor reduce search email valid

ddikman:
n/a

### Change description ###
Reused validation code. I think it was put in everywhere due to the use of a `setTimeout` in the method, which probably made it not function correctly.

```
[ ] Yes
[X] No
```


AdminWebsite/AdminWebsite/ClientApp/src/app/booking/add-participant/add-participant.component.html
```diff
@@ -23,7 +23,7 @@ <h2 class="govuk-error-summary__title" id="error-summary-title">
           <li *ngIf="role.invalid || !isRoleSelected">
             <a (click)="goToDiv('role-error')" href="javascript: void(0)">{{constants.Error.RoleErrorMsg}}</a>
           </li>
-          <li *ngIf="emailInvalid()">
+          <li *ngIf="validEmail()">
```
rehmanab: Should this be inverted ?

ddikman: Good catch 🙌 I've inverted this

PR #116, opened 2019-07-09
Feature/fix questionnaire duplications 4691

valentinabennett:
VIH-4691

### Change description ###
Fix questionnaire duplications

```
[ ] Yes
[x ] No
```

### DIFF
--- a/AdminWebsite/AdminWebsite/ClientApp/src/app/suitability/services/questionnaire.service.ts
+++ b/AdminWebsite/AdminWebsite/ClientApp/src/app/suitability/services/questionnaire.service.ts
@@ -19,10 +19,8 @@ export class QuestionnaireService {
 
   constructor(private service: ScrollableSuitabilityAnswersService) { }
 
-  async loadNext(hasMore: boolean, nextCursor: string): Promise<QuestionnaireResponses> {
-    if (!hasMore) {
-      return new QuestionnaireResponses([], false, '');
-    }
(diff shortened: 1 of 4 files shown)


AdminWebsite/AdminWebsite/ClientApp/src/app/suitability/services/questionnaire.service.ts
```diff
@@ -5,32 +5,31 @@ import { ScrollableSuitabilityAnswersService } from './scrollable-suitability-an
...
-  private nextCursor = '';
-  private hasMore = true;
 
   constructor(private service: ScrollableSuitabilityAnswersService) { }
 
-  async loadNext(): Promise<QuestionnaireResponses> {
-    if (!this.hasMore) {
-      return new QuestionnaireResponses([], false);
+  async loadNext(hasMore: boolean, nextCursor: string): Promise<QuestionnaireResponses> {
+    if (!hasMore) {
```
ddikman: It looks like `hasMore` is now handled outside of this class so maybe we can move that out to the caller or is there a reason to have it inside here?

valentinabennett: Yes, I think we don't need it any more in the service. I moved next cursor and hasMore to component as it was a reason for duplication, service was keeping value of this variables and don't care about component actions.

PR #454, opened 2020-09-23
VIH-6550 Assign defence advocate to professional user group

shaed-parkar:

VIH-6550

### Change description ###

Assign defence advocate to professional user group

```
[ ] Yes
[x] No
```

### DIFF
--- a/AdminWebsite/AdminWebsite/Services/UserAccountService.cs
+++ b/AdminWebsite/AdminWebsite/Services/UserAccountService.cs
@@ -25,7 +25,7 @@ namespace AdminWebsite.Services
         /// </summary>
         /// <param name="participant"></param>
         /// <returns></returns>
-        Task UpdateParticipantUsername(ParticipantRequest participant);
+        Task<string> UpdateParticipantUsername(ParticipantRequest participant);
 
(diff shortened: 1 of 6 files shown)


AdminWebsite/AdminWebsite/Services/UserAccountService.cs
```diff
@@ -129,7 +130,7 @@ private async Task<NewUserResponse> CreateNewUserInAD(ParticipantRequest partici
 
             await _userApiClient.AddUserToGroupAsync(addUserToGroupRequest);
 
-            if (participant.Hearing_role_name == REPRESENTATIVE)
+            if (participant.Hearing_role_name == REPRESENTATIVE || participant.Hearing_role_name == DEFENCE_ADVOCATE)
```
RuwanikaWeerakkody, line 133: @shaed-parkar  could we go by the user role here rather than adding all the hearing roles here. Hearing roles are updated often and we don't really want to change the user account service based on this.

shaed-parkar: we don't know the user role by this point unfortunately. I tried but the user role is determined by the bookings api once created

RocheL (approved):

PR #472, opened 2020-10-12
VIH-6604 Fix for multi dates hearing

valentinabennett:

https://tools.hmcts.net/jira/browse/VIH-6604
### Change description ###

Fix multi day hearing

```
[X ] Yes
[ ] No
```

### DIFF
--- a/AdminWebsite/AdminWebsite/Helper/DateListMapper.cs
+++ b/AdminWebsite/AdminWebsite/Helper/DateListMapper.cs
@@ -21,10 +21,11 @@ namespace AdminWebsite.Helper
 
         public static List<DateTime> GetListOfWorkingDates(DateTime startDate, DateTime endDate)
         {
+
             var dates = new List<DateTime>();
             var nextDate = startDate.AddDays(1);
 
-            for (var dt = nextDate; dt <= endDate; dt = dt.AddDays(1))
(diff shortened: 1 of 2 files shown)


AdminWebsite/AdminWebsite/Helper/DateListMapper.cs
```diff
@@ -21,10 +21,11 @@ public static List<DateTime> GetListOfWorkingDates(string startDate, string endD
 
         public static List<DateTime> GetListOfWorkingDates(DateTime startDate, DateTime endDate)
         {
+
             var dates = new List<DateTime>();
             var nextDate = startDate.AddDays(1);
 
-            for (var dt = nextDate; dt <= endDate; dt = dt.AddDays(1))
+            for (var dt = nextDate; dt.Date <= endDate.Date; dt = dt.AddDays(1))
```
RuwanikaWeerakkody, line 28: Should be both start and end date inclusive. No need to add a day to the end date

RuwanikaWeerakkody (changes requested):

AdminWebsite/AdminWebsite/Helper/DateListMapper.cs
valentinabennett: The first booking is done in normal way, so I pass only clones dates, it should be like that

RuwanikaWeerakkody (approved):

PR #484, opened 2020-10-19
VIH-6275 include users in AD in search when deleting

shaed-parkar:

VIH-6275

### Change description ###

* also search in AD when searching for participants to delete
* on create hearing, explicitly search for contact before creating an account for new participants

```
[ ] Yes
[x] No
```

### DIFF
--- a/AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
+++ b/AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
@@ -110,28 +110,15 @@ namespace AdminWebsite.Controllers
                 string adUserId;
                 if(string.IsNullOrWhiteSpace(participant.Username))
                 {
-                    // check user contact email already exist
-                    var user = await _userAccountService.GetUserByContactEmail(participant.Contact_email);
-                    if (user != null)
-                    {
-                        participant.Username = user.User_name;
-                        adUserId = user.User_id;
-                    }
-                    else
-                    {
-                        // create user
-                        var newUser = await _userAccountService.CreateNewUserInAD(participant);
-                        participant.Username = newUser.Username;
-                        adUserId = newUser.User_id;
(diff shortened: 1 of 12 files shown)


AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
```diff
@@ -110,18 +110,15 @@ public async Task<ActionResult<HearingDetailsResponse>> Post([FromBody] BookNewH
...
+                    adUserId = await _userAccountService.UpdateParticipantUsername(participant);
                 }
                 else
                 {
                     // get user
                     adUserId = await _userAccountService.GetAdUserIdForUsername(participant.Username);
                 }
-                    
-                usernameAdIdDict[participant.Username] = adUserId;
+                // username's participant will be set by this point
```
rehmanab, line 120: No need for comment Me thinks

AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
```diff
@@ -110,18 +110,15 @@ public async Task<ActionResult<HearingDetailsResponse>> Post([FromBody] BookNewH
...
+                usernameAdIdDict[participant.Username!] = adUserId;
```
rehmanab, line 121: No need for Null forgiving here as a null string will throw when indexing into a dic

rehmanab (approved):
A couple of non urgent comments

AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
shaed-parkar: The IDE kept raising warning here, I forgave it because the compiler could not recognise the value would be set by useraccount service

valentinabennett (approved):

PR #557, opened 2021-01-12
VIH-6894 Adding retry to the AAD user adding to group

rehmanab:
https://tools.hmcts.net/jira/browse/VIH-6894


shaed-parkar (approved):
LGTM

PR #575, opened 2021-02-02
VIH-6922 connection service

phillipjchambers:

VIH-6922

### Change description ###

Add a connection service which periodically tests that the user can make web requests.

- add ConnectionService service, injected in root
- subscribe to connection status in app component & redirect to ErrorComponent on connection error
- modify ErrorComponent wording as per copy provided (see ticket)
- display the correct error for the current connection status
- update HeaderComponent to hide navigation when there is not connection
- update all existing tests that now depend on ConnectionService
- add new tests for AppComponent & ConnectionService

```
[ ] Yes
[x] No
```


AdminWebsite/AdminWebsite/ClientApp/src/app/app.module.ts
```diff
@@ -94,7 +94,7 @@ export function getSettings(configService: ConfigService) {
         ErrorService,
         PageTrackerService,
         AppInsightsLogger,
-        WindowRef
+        WindowRef,
```
shaed-parkar, line 97: You might want to check you're using the same linting or prettier rules. I didn't expect this to change.

AdminWebsite/AdminWebsite/ClientApp/src/app/error/error.component.html
```diff
@@ -1,6 +1,31 @@
 <div class="govuk-grid-row">
-  <div class="govuk-grid-column-full">
-    <h1 class="govuk-heading-xl">Sorry, there is a problem with the service</h1>
-    <p class="govuk-body">Please try again later.</p>
-  </div>
+  <ng-container *ngIf="hasConnection$ | async">
```
shaed-parkar, line 2: Could `*ngIf="hasConnection$ | async;noConnection"` work here?

shaed-parkar (approved):
Nice work! I really like how you've implemented the connection service.

AdminWebsite/AdminWebsite/ClientApp/src/app/error/error.component.html
phillipjchambers: Using a ngTemplate? Probably, yeah - you might need to use parens, though, like this `*ngIf="(hasConnection$ | async); else noConnection"`

AdminWebsite/AdminWebsite/ClientApp/src/app/app.module.ts
phillipjchambers: Good point, will remove

shaed-parkar (approved):

shaed-parkar (approved):

PR #602, opened 2021-02-17
VIH-7320 change how representatives are determined

phillipjchambers:

VIH-7320

### Change description ###

Consider a participant a representative if the have a representee instead of based on their hearing role name. Currently, a participant is only considered a representative if they have the hearing_role_name "Representative", which causes an incorrect UI for barristers & other reps.

```
[ ] Yes
[x] No
```

### DIFF
--- a/AdminWebsite/AdminWebsite/ClientApp/src/app/booking/participants-list/participants-list.component.html
+++ b/AdminWebsite/AdminWebsite/ClientApp/src/app/booking/participants-list/participants-list.component.html
@@ -1,8 +1,8 @@
 <div id="roundedborder">
   <h3 class="govuk-heading-m">{{ isEditMode ? 'Participants' : 'Participants added' }}</h3>
   <div class="vhtable" id="participants-list">
-    <div *ngFor="let pat of participants">
-      <div *ngIf="pat.is_judge">
+    <div *ngFor="let p of participants">
+      <div *ngIf="p.is_judge">
         <div class="govuk-grid-row vh-row">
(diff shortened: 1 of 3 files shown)


AdminWebsite/AdminWebsite/ClientApp/src/app/booking/participants-list/participants-list.component.html
```diff
@@ -1,8 +1,8 @@
 <div id="roundedborder">
   <h3 class="govuk-heading-m">{{ isEditMode ? 'Participants' : 'Participants added' }}</h3>
   <div class="vhtable" id="participants-list">
-    <div *ngFor="let pat of participants">
```
RuwanikaWeerakkody, line 4: Hi Phil, I don't think we should do this change. UI should use the hearing role representative. So the current logic is correct

RuwanikaWeerakkody (changes requested):

AdminWebsite/AdminWebsite/ClientApp/src/app/booking/participants-list/participants-list.component.html
phillipjchambers: Even though it doesn't include Barristers, etc?

RuwanikaWeerakkody: Looks like those migrations could be wrong then Phill. Should they have been added as rep's?

PR #614, opened 2021-02-24
VIH-7363 for cacd we are showing 'none' for party in participants added list

RocheL:
https://tools.hmcts.net/jira/browse/VIH-7363
### Change description ###
 - fix for case roles
```
[ ] Yes
[x] No
```

### DIFF
--- a/AdminWebsite/AdminWebsite/ClientApp/src/app/booking/participants-list/participants-list.component.ts
+++ b/AdminWebsite/AdminWebsite/ClientApp/src/app/booking/participants-list/participants-list.component.ts
@@ -99,4 +99,11 @@ export class ParticipantsListComponent implements OnInit, OnChanges {
         }
         return interpretee;
     }
+    getCaseName(particpant: ParticipantModel): string {
+        return particpant.case_role_name.toLowerCase() === 'none' ||
+            particpant.case_role_name.toLowerCase() === 'observer' ||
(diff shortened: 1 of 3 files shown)


AdminWebsite/AdminWebsite/ClientApp/src/app/booking/participants-list/participants-list.component.ts
```diff
@@ -99,4 +99,11 @@ export class ParticipantsListComponent implements OnInit, OnChanges {
         }
         return interpretee;
     }
+    getCaseName(particpant: ParticipantModel): string {
```
phillipjchambers, line 102: I think you should use an array here & includes, like this:
```
['none','observer','panel member'].includes(participant.case_role_name)
```

phillipjchambers: Having said that, I've made some changes in this same area so the point might be moot

phillipjchambers: But, generally, when you're doing multiple string comparisons on the same prop, Array.includes is nice

RocheL: yes, will merge with the refactored version.

PR #643, opened 2021-03-17
VIH-7312 Enable judge multiday emails

virensareen:
VIH-7312

### Change description ###
Enable judge multiday emails

```
[ ] Yes
[x] No
```

### DIFF
--- a/AdminWebsite/AdminWebsite.UnitTests/Services/HearingServiceTests.cs
+++ b/AdminWebsite/AdminWebsite.UnitTests/Services/HearingServiceTests.cs
@@ -97,22 +97,6 @@ namespace AdminWebsite.UnitTests.Services
         }
         
         [Test]
-        public async Task should_not_create_notification_for_judge_email_when_groupid_is_null()
-        {
-            var judge = _hearing.Participants.First(x => x.User_role_name == "Judge");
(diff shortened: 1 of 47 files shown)


AdminWebsite/AdminWebsite.UnitTests/Services/HearingServiceTests.cs
```diff
@@ -95,6 +95,22 @@ public async Task should_send_multiday_confirmation_email_to_all_participants_ex
                     x => x.CreateNewNotificationAsync(It.Is<AddNotificationRequest>(r => r.ParticipantId != judge.Id)),
                     Times.Exactly(_hearing.Participants.Count(x => x.User_role_name.ToLower() != "judge")));
         }
+        
+        [Test]
+        public async Task should_not_create_notification_for_judge_email_when_groupid_is_null()
```
shaed-parkar, line 100: Group Id will never be null. You can remove this test. If necessary, update the builders so it's not null and set to be the same value as the hearing ID unless it's representing a cloned hearing

shaed-parkar (changes requested):
Looks good, you need to remove tests for scenarios that shouldn't exist

RuwanikaWeerakkody (approved):

shaed-parkar (approved):

PR #647, opened 2021-03-18
VIH-7472 removed null text entry

surfsquid:
https://tools.hmcts.net/jira/browse/VIH-7472

 removed null text entry

```
[ ] Yes
[X] No
```


AdminWebsite/AdminWebsite/ClientApp/src/app/services/pipe-stringifier.service.ts
```diff
@@ -19,9 +19,13 @@ export class PipeStringifierService {
     decode<T>(input: string): T {
         const output = {};
         const keyValuePairs = input.match(/[^|]+\|[^|]+/g);
-        keyValuePairs.forEach(property => {
+        keyValuePairs?.forEach(property => {
             const pair = property.split('|');
-            output[pair[0]] = pair[1];
+            if (pair[1]) {
```
levibutler, line 24: would this work?
`output[pair[0]] =pair[1] ?? '';`

levibutler (approved):

PR #649, opened 2021-03-20
VIH-7430 judge email phone in booking details

virensareen:
VIH-7430

### Change description ###
Display judge email and phone in booking details

```
[ ] Yes
[x] No
```


RuwanikaWeerakkody (approved):

AdminWebsite/AdminWebsite.AcceptanceTests/Steps/BookingDetailsSteps.cs
```diff
@@ -98,10 +99,46 @@ private void VerifyJudgeInParticipantsList()
...
+                return null;
+            }
+            return email;
+        }
+        
+        private static OtherInformationDetails GetOtherInformationObject(string otherInformation)
+        {
+            try
+            {
+                var properties = otherInformation.Split("|");
```
levibutler, line 127: tbc it adds empty || for null values?

virensareen: Null or blank values have an empty string yes

AdminWebsite/AdminWebsite.AcceptanceTests/Steps/BookingDetailsSteps.cs
```diff
@@ -98,10 +99,46 @@ private void VerifyJudgeInParticipantsList()
...
+                {
+                    JudgeEmail = properties[2],
+                    JudgePhone = properties[4],
+                    OtherInformation = properties[6]
+                };
+            }
+            catch (Exception)
+            {
+                var properties = otherInformation.Split("|");
+                return new OtherInformationDetails {OtherInformation = properties[2]};
```
levibutler, line 138: where did 2 come from? likely to come here because its not using this new format and doesnt contain any `|` ? properties[2] would just throw another our of range exception?

virensareen: The old format would be parsed to `|OtherInformation|Sometext|`, so:
[0] = empty string
[1] = OtherInformation
[2] = Sometext --> We want this value

AdminWebsite/AdminWebsite/Extensions/HearingDetailsResponseExtensions.cs
```diff
@@ -90,7 +90,13 @@ private static OtherInformationDetails GetOtherInformationObject(string otherInf
...
             catch (Exception)
             {
-                return new OtherInformationDetails {OtherInformation = otherInformation};
+                var properties = otherInformation.Split("|");
+                if (properties.Length > 1)
+                {
+                    return new OtherInformationDetails {OtherInformation = properties[2]};
+                }
+
+                return new OtherInformationDetails {OtherInformation = properties[0]};
```
levibutler, line 99: same as = otherInformation, this might be clearer tho?

AdminWebsite/AdminWebsite/Extensions/HearingDetailsResponseExtensions.cs
```diff
@@ -90,7 +90,13 @@ private static OtherInformationDetails GetOtherInformationObject(string otherInf
             }
             catch (Exception)
             {
-                return new OtherInformationDetails {OtherInformation = otherInformation};
+                var properties = otherInformation.Split("|");
+                if (properties.Length > 1)
```
levibutler, line 94: should this be > 2

AdminWebsite/AdminWebsite/Extensions/HearingDetailsResponseExtensions.cs
virensareen: Do you mean the parameter? That is a pipe separated string. To map the correct values to the object, we need to split the string into an array, then extract the corresponding values

AdminWebsite/AdminWebsite/Extensions/HearingDetailsResponseExtensions.cs
virensareen: We just care if the length is 2 or higher? Because we want the 3rd element.

levibutler (commented):
cool, couple comments

AdminWebsite/AdminWebsite/Extensions/HearingDetailsResponseExtensions.cs
levibutler: length doesnt start from 0 but index does

AdminWebsite/AdminWebsite/Extensions/HearingDetailsResponseExtensions.cs
levibutler: if it was pipe separated it would go inside the block above

AdminWebsite/AdminWebsite.AcceptanceTests/Steps/BookingDetailsSteps.cs
levibutler: oh, is there a migration to change the old information to this format?

AdminWebsite/AdminWebsite.AcceptanceTests/Steps/BookingDetailsSteps.cs
levibutler: super

AdminWebsite/AdminWebsite.AcceptanceTests/Steps/BookingDetailsSteps.cs
virensareen: I'm unsure, I think that may have been part of David's work? This is the format I was getting back whilst testing

AdminWebsite/AdminWebsite/Extensions/HearingDetailsResponseExtensions.cs
virensareen: We can still get a pipe separated format when we have old data (ie no judge email/phone). The old data looks like this now `|OtherInformation|Sometext|`, this was picked up in testing. To cater for this, we need this if statement, otherwise the smoke test fails

AdminWebsite/AdminWebsite.AcceptanceTests/Steps/BookingDetailsSteps.cs
levibutler: can we add a length check like the other to be safe?

AdminWebsite/AdminWebsite/Extensions/HearingDetailsResponseExtensions.cs
levibutler: super, comment still stands tho, line 96 will set to [2], this one here is setting to `otherinformation` if it hasnt split as expected

virensareen: But if we do that the property within the object will equal `|OtherInformation|Sometext|`, we just want it to equal `Sometext`

AdminWebsite/AdminWebsite/ClientApp/src/app/bookings-list/participant-details/participant-details.component.html
```diff
@@ -79,11 +85,16 @@
...
-      <div class="govuk-grid-row vh-mtop" *ngIf="participant?.HearingRoleName !== 'Judge'">
+      <div class="govuk-grid-row vh-mtop">
         <div class="govuk-grid-column-one-third contact-label">Telephone</div>
         <div id="participant-{{ participant?.ParticipantId }}-phone" class="govuk-grid-column-two-thirds wrap">
-          {{ participant?.Phone }}
+          <div *ngIf="!isJudge(participant)">{{ participant?.Phone }}</div>
+          <div *ngIf="isJudge(participant) && (getJudgePhone() !== '' || getJudgeEmail() !== undefined || getJudgeEmail() !== null)">
+            {{ getJudgePhone() }}
+          </div>
+          <div *ngIf="isJudge(participant) && (getJudgePhone() === '' || getJudgeEmail() === undefined || getJudgeEmail() === null)">
```
levibutler, line 95: should this be judgePhone? and the ones above be email?
also this is quite long, maybe we could move to ts and create a get with a nice description? and we could test

AdminWebsite/AdminWebsite/Extensions/HearingDetailsResponseExtensions.cs
levibutler: 🙄 think we might be best to have a quick chat.. dont think were understanding each other here :)

AdminWebsite/AdminWebsite/ClientApp/src/app/bookings-list/participant-details/participant-details.component.html
```diff
@@ -70,7 +70,13 @@
       <div class="govuk-grid-row">
         <div class="govuk-grid-column-one-third contact-label">Email</div>
         <div id="participant-{{ participant?.ParticipantId }}-email" class="govuk-grid-column-two-thirds wrap">
-          {{ participant?.Email }}
+          <div *ngIf="!isJudge(participant)">{{ participant?.Email }}</div>
+          <div *ngIf="isJudge(participant) && (getJudgeEmail() !== '' || getJudgeEmail() === undefined || getJudgeEmail() !== null)">
```
phillipjchambers, line 74: Checking for an empty string, undefined and null can be done in one op - they're all falsy. Instead of checking that the return is no falsy, I recommend just checking it is truthy, like this

```
isJudge(participant) && getJudgeEmail()
```

I would also suggest making `getJudgeEmail` a getter because (a) it has _get_ in its name and (b) it takes no params

Falsy ref: https://developer.mozilla.org/en-US/docs/Glossary/Falsy

AdminWebsite/AdminWebsite/ClientApp/src/app/bookings-list/participant-details/participant-details.component.html
```diff
@@ -70,7 +70,13 @@
...
+            {{ getJudgeEmail() }}
+          </div>
+          <div *ngIf="isJudge(participant) && (getJudgeEmail() === '' || getJudgeEmail() === undefined || getJudgeEmail() === null)">
```
phillipjchambers, line 77: Here you can just do the opposite of above, we're checking they are falsy now, so try this:
```
isJudge(participant) || !getJudgeEmail()
```

Having said all of that, it's sometimes neater to use templates instead of mutliple ngIfs, like this:

```
<div *ngIf="isJudge(participant); else participantEmail">
  <div *ngIf="judgeEmail; else judgeNoEmail">
    {{ judgeEmail }}
  </div>
  <ng-template #judgeNoEmail>
    <div>
        TBC
    </div>
  </ng-template>
</div>

<ng-template #participantEmail>
  <div>
    {{ participant.Email }}
  </div>
</ng-template>
```

What's happening:
1. The else on the ngIf refers to a nested ng-template, which is denoted by the hash: `#judgeNoEmail`. This template will be displayed if the `ngIf` returns false.
2. I've inverted your logic to check for the judge first
3. I'm assuming the `getJudgeEmail()` is now a getter, so it becomes just `judgeEmail`
4. I've left all the divs in because I'm assuming (without checking) that they're needed for structure, i.e., block level. However, if they're being displayed inline anyway, you can remove any `<div>` and replace any divs that contain ngIfs with ng-containers (as these don't get rendered to the DOM so they keep the doc neater.

AdminWebsite/AdminWebsite/ClientApp/src/app/bookings-list/participant-details/participant-details.component.html
```diff
@@ -79,11 +85,16 @@
           {{ participant?.UserName }}
         </div>
       </div>
-
-      <div class="govuk-grid-row vh-mtop" *ngIf="participant?.HearingRoleName !== 'Judge'">
+      <div class="govuk-grid-row vh-mtop">
         <div class="govuk-grid-column-one-third contact-label">Telephone</div>
         <div id="participant-{{ participant?.ParticipantId }}-phone" class="govuk-grid-column-two-thirds wrap">
-          {{ participant?.Phone }}
+          <div *ngIf="!isJudge(participant)">{{ participant?.Phone }}</div>
```
phillipjchambers, line 91: See above

AdminWebsite/AdminWebsite/ClientApp/src/app/bookings-list/participant-details/participant-details.component.ts
```diff
@@ -9,9 +11,24 @@ import { ParticipantDetailsModel } from '../../common/model/participant-details.
...
-
+    @Input()
+    hearing: BookingsDetailsModel;
     @Input()
     vh_officer_admin: boolean;
 
     constructor() {}
+
+    getJudgeEmail() {
+        const email = OtherInformationModel.init(this.hearing.OtherInformation).JudgeEmail;
```
phillipjchambers, line 22: Suggest
```ts
get judgeEmail() {
  return OtherInformationModel.init(this.hearing.OtherInformation).JudgeEmail;
}
```

AdminWebsite/AdminWebsite/ClientApp/src/app/bookings-list/participant-details/participant-details.component.ts
```diff
@@ -9,9 +11,24 @@ import { ParticipantDetailsModel } from '../../common/model/participant-details.
...
+        return email;
+    }
+
+    getJudgePhone() {
```
phillipjchambers, line 26: As above

AdminWebsite/AdminWebsite/ClientApp/src/app/bookings-list/participant-details/participant-details.component.ts
```diff
@@ -9,9 +11,24 @@ import { ParticipantDetailsModel } from '../../common/model/participant-details.
...
+        const phone = OtherInformationModel.init(this.hearing.OtherInformation).JudgePhone;
+        return phone;
+    }
+
+    isJudge(participant: ParticipantDetailsModel) {
+        return participant.isJudge;
```
phillipjchambers, line 32: This exists on the model (I can see it below) - is it needed here?

levibutler (dismissed):

phillipjchambers (dismissed):
LGTM

phillipjchambers (approved):

PR #670, opened 2021-03-30
VIH-7541 stop bookings on public holidays

shaed-parkar:

VIH-7541 

### Change description ###

* stop users from booking hearings on public holidays in Wales and England
* skip public holidays as well as weekends when using MDH bookings

image

```
[ ] Yes
[x] No
```


AdminWebsite/AdminWebsite.UnitTests/Helper/DateListMapperTest.cs
```diff
@@ -31,5 +33,23 @@ public void Should_return_empty_list_if_end_and_start_dates_is_the_same()
 
             result.Count.Should().Be(expectDays);
         }
+        
+        [Test]
+        public void Should_return_range_of_dates_not_included_weekends_and_public_holidays()
+        {
+            var startDate = new DateTime(2020, 10, 1, 4, 30, 0, 0);
```
levibutler, line 40: should we have tests for more edge case ranges?
same start date end date, 
start date or end date on weekend
end date larger than start date 💥

AdminWebsite/AdminWebsite/ClientApp/src/app/common/custom-validations/public-holiday-validator.ts
```diff
@@ -0,0 +1,17 @@
+import { AbstractControl, ValidatorFn } from '@angular/forms';
+
+export function notPublicHolidayDateValidator(publicHolidays: Date[]): ValidatorFn {
+    return (ctrl: AbstractControl) => {
+        if (!publicHolidays.length) {
+            return null;
+        }
+        const pbTimes = publicHolidays.map(x => x.getTime());
```
levibutler, line 8: just checking getTime returns a date? or i guess they would all be midnight and it would always be a holiday?

AdminWebsite/AdminWebsite/ClientApp/src/app/services/reference-data.service.ts
```diff
@@ -1,14 +1,25 @@
...
+    private publicHolidays: PublicHolidayResponse[];
     constructor(private bhClient: BHClient) {}
 
     getCourts(): Observable<HearingVenueResponse[]> {
         return this.bhClient.getCourts();
     }
+
+    fetchPublicHolidays(): void {
+        this.bhClient.publicHolidays().subscribe({
+            next: pb => (this.publicHolidays = pb)
```
levibutler, line 18: do we need next? ive seen it complain before too saying its obsolete but i thought it was because of passing in the object, i think we can do
`this.bhClient.publicHolidays().subscribe(pb => (this.publicHolidays = pb));`
would do the same and not going obsolete? that i know of? i could ofc be wrong about the whole thing

AdminWebsite/AdminWebsite/Contracts/Responses/PublicHolidayResponse.cs
```diff
@@ -0,0 +1,20 @@
...
+{
+    /// <summary>
+    /// A public holiday
+    /// </summary>
+    public class PublicHolidayResponse
+    {
+        /// <summary>
+        /// Name of a public holiday
+        /// </summary>
+        public string Name { get; set; }
```
levibutler, line 13: some holidays are specific to some regions, we support welsh now, just checking this has been discussed at some point

AdminWebsite/AdminWebsite/ClientApp/src/app/common/custom-validations/public-holiday-validator.ts
shaed-parkar: getTime is like timespan. the equality operator doesn't compare dates in the way we would in c#

AdminWebsite/AdminWebsite/Services/Models/PublicHoliday.cs
```diff
@@ -0,0 +1,31 @@
+using System;
+using System.Collections.Generic;
+using Newtonsoft.Json;
+
+namespace AdminWebsite.Services.Models
+{
+    class UkHolidaysResponse
```
levibutler, line 7: spoke too soon :P

AdminWebsite/AdminWebsite/ClientApp/src/app/services/reference-data.service.ts
shaed-parkar: i got hit pretty hard in a previous PR not following this else I usually do so like you suggested

AdminWebsite/AdminWebsite/Contracts/Responses/PublicHolidayResponse.cs
shaed-parkar: The name gets pulled in from the GOVUK API, I pull in the names they provide which include England and Wales

AdminWebsite/AdminWebsite/Services/UkPublicHolidayRetriever.cs
```diff
@@ -0,0 +1,53 @@
...
+        
+        private async Task<List<PublicHoliday>> GetHolidaysFromApi()
+        {
+            var holidays = new List<PublicHoliday>();
+            var ukHolidaysUri = @"https://www.gov.uk/bank-holidays.json";
+            
+            var response = await _httpClient.GetAsync(ukHolidaysUri);
+            response.EnsureSuccessStatusCode();
+            
+            if (!response.IsSuccessStatusCode) return holidays;
```
levibutler, line 42: maybe return on new line with some `{}` but ive seen a mix here so maybe no need 🤷‍♂️ just personal preference

AdminWebsite/AdminWebsite.UnitTests/Helper/DateListMapperTest.cs
shaed-parkar: That already exists in another test :)

levibutler (approved):
LGTM

AdminWebsite/AdminWebsite/Services/UkPublicHolidayRetriever.cs
shaed-parkar: I'm actually a fan of single line returns like this 😅

PR #672, opened 2021-04-07
VIH-7476 Logging Debug changed to Logging Information

ghost:
https://tools.hmcts.net/jira/browse/VIH-7476

### Change description ###
Swapping calls order and updating logging.

```
[ ] Yes
[x] No
```

### DIFF
--- a/AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
+++ b/AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
@@ -85,17 +85,17 @@ namespace AdminWebsite.Controllers
 
                 newBookingRequest.Created_by = _userIdentity.GetUserIdentityName();
 
-                _logger.LogDebug("BookNewHearing - Attempting to send booking request to Booking API");
+                _logger.LogInformation("BookNewHearing - Attempting to send booking request to Booking API");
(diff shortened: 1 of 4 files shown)


shaed-parkar (changes requested):
No need to change the logging level for each type. We can pick this story up again together, but this is actually missing the fix for the original issue in the ticket

AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
```diff
@@ -306,13 +306,13 @@ private async Task SendJudgeEmailIfNeeded(HearingDetailsResponse updatedHearing,
         {
             if (newParticipantList.Any())
             {
-                _logger.LogDebug("Sending email notification to the participants");
+                _logger.LogInformation("Sending email notification to the participants");
```
RuwanikaWeerakkody, line 309: Log levels in PROD may not log LogInformation. Please check the log levels in PROD first.

shaed-parkar:
Also please move the send email before the group assignment as discussed

AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
shaed-parkar: AppInsights will capture info and higher :)

shaed-parkar (approved):

RuwanikaWeerakkody (approved):

PR #693, opened 2021-04-30
VIH-7621 Making a user unable to edit a confirmed hearing under 30 minutes

mulvaneyjHMCTS:
VIH-7621

### Change description ###
The hearing scheduled time was being reset to the end of the day
30 minutes were being subtracted to the current time rather than added

```
[ ] Yes
[x] No
```

### DIFF
--- a/AdminWebsite/AdminWebsite/ClientApp/src/app/bookings-list/booking-details/booking-details.component.ts
+++ b/AdminWebsite/AdminWebsite/ClientApp/src/app/bookings-list/booking-details/booking-details.component.ts
@@ -104,7 +104,7 @@ export class BookingDetailsComponent implements OnInit, OnDestroy {
 
     setTimeObserver() {
         if (this.booking) {
-            const endofday = this.booking.scheduled_date_time;
+            const endofday = new Date(this.booking.scheduled_date_time);
             endofday.setHours(23, 59);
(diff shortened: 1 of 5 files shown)


AdminWebsite/AdminWebsite/ClientApp/src/app/bookings-list/booking-details/booking-details.component.ts
```diff
@@ -104,7 +104,7 @@ export class BookingDetailsComponent implements OnInit, OnDestroy {
 
     setTimeObserver() {
         if (this.booking) {
-            const endofday = this.booking.scheduled_date_time;
+            const endofday = new Date(this.booking.scheduled_date_time);
```
phillipjchambers, line 107: Was this change put in because `this.booking.scheduled_date_time` wasn't always a Date? Might be worth checking it's truthy before doing this, if there's a chance it can be undefined etc

mulvaneyjHMCTS: No this was because the scheduled_date_time was being linked to the endofday const rather than just taking the value so anytime it was updated the scheduled_date_time would be updated which is not what we want. The new Date solved this

kevin-hmcts (approved):

PR #695, opened 2021-04-30
VIH-7629 Other Information Changes

FarronF:
VIH-7629

### Change description ###
Only send other information details to API if they are present. Without this change the requests work for the latest in master but on dev return a 403, see ticket for details.

```
[ ] Yes
[x] No
```

### DIFF
--- a/AdminWebsite/AdminWebsite/ClientApp/src/app/common/model/other-information.model.ts
+++ b/AdminWebsite/AdminWebsite/ClientApp/src/app/common/model/other-information.model.ts
@@ -6,10 +6,19 @@ export class OtherInformationModel {
     OtherInformation?: string;
 
     static init(otherInformation: string): OtherInformationModel {
-        let otherInfo: OtherInformationModel = {};
+        let otherInfo: OtherInformationModel = new OtherInformationModel();
         const stringifier = new PipeStringifierService();
         try {
             otherInfo = stringifier.decode(otherInformation);
+            if (!otherInfo.JudgeEmail) {
+                otherInfo.JudgeEmail = null;
+            }
+            if (!otherInfo.JudgePhone) {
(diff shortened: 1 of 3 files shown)


AdminWebsite/AdminWebsite/ClientApp/src/app/common/model/other-information.model.ts
```diff
@@ -6,10 +6,19 @@ export class OtherInformationModel {
...
             otherInfo = stringifier.decode(otherInformation);
+            if (!otherInfo.JudgeEmail) {
+                otherInfo.JudgeEmail = null;
+            }
+            if (!otherInfo.JudgePhone) {
+                otherInfo.JudgePhone = null;
+            }
+            if (!otherInfo.OtherInformation) {
+                otherInfo.OtherInformation = null;
+            }
```
FarronF, line 21: Probably a nicer way to do this but just ensures that properties are on the object returned previously would return {}

AdminWebsite/AdminWebsite/ClientApp/src/app/services/pipe-stringifier.service.ts
```diff
@@ -9,7 +9,7 @@ export class PipeStringifierService {
     encode<T>(input: T): string {
         let output = '';
         for (const property in input) {
-            if (Object.prototype.hasOwnProperty.call(input, property)) {
+            if (Object.prototype.hasOwnProperty.call(input, property) && !!input[property]) {
```
FarronF, line 12: Dev Api seems to not like getting the following
|JudgeEmail||JudgePhone||OtherInformation|something

This would mean that would return
|OtherInformation|something

kevin-hmcts (approved):

PR #706, opened 2021-05-11
VIH-7556 vh officer adds participant to booking within 30mins

mulvaneyjHMCTS:
VIH-7556

### Change description ###
Vh officer adds participant to booking within 30mins
Updated the way edit and cancel buttons appear

```
[ ] Yes
[x] No
```


AdminWebsite/AdminWebsite/ClientApp/src/app/booking/breadcrumb/breadcrumbItem.model.ts
```diff
@@ -1,14 +1,16 @@
 export class BreadcrumbItemModel {
```
phillipjchambers, line 1: I know this isn't your change, so no need to do anything but for future reference: this class can just be an interface. The ctor is pointless.

AdminWebsite/AdminWebsite/ClientApp/src/app/booking/participant/list/participant-list.component.ts
```diff
@@ -109,4 +104,20 @@ export class ParticipantListComponent implements OnInit, OnChanges {
             i.is_interpretee = false;
         });
     }
+
+    canEditParticipant(particpant: ParticipantModel): boolean {
```
phillipjchambers, line 108: Can this be simplified?

AdminWebsite/AdminWebsite/ClientApp/src/app/common/model/participant.model.ts
```diff
@@ -23,6 +23,7 @@ export class ParticipantModel {
     is_interpretee?: boolean | undefined;
     user_role_name?: string | undefined;
     is_courtroom_account?: boolean;
+    addedDuringHearing?: boolean | undefined;
```
phillipjchambers, line 26: You don't need a union type here - it's already marked as optional with the `?` so the union is unecessary

AdminWebsite/AdminWebsite/ClientApp/src/app/services/video-hearings.service.ts
```diff
@@ -441,4 +443,20 @@ export class VideoHearingsService {
     getJudge(): ParticipantModel {
         return this.modelHearing.participants.find(x => x.is_judge);
     }
+
+    isConferenceClosed(): boolean {
+        this.checkForExistingHearing();
```
phillipjchambers, line 448: Should this be asynchronous? i.e., use an obseravble?

AdminWebsite/AdminWebsite/ClientApp/src/app/services/video-hearings.service.ts
```diff
@@ -441,4 +443,20 @@ export class VideoHearingsService {
...
+        return this.modelHearing.status === BookingStatus.Created && this.modelHearing.telephone_conference_id === '';
+    }
+
+    isHearingAboutToStart(): boolean {
+        this.checkForExistingHearing();
```
phillipjchambers, line 453: As above

phillipjchambers: Just because I think the following lines rely on whatever ^ that does completing...but I could be wrong

AdminWebsite/AdminWebsite/ClientApp/src/app/services/video-hearings.service.ts
FarronF: This is getting details from session storage so it's all local. I would like to see if there's a way this could be done once instead of every time either of these are called.

There's the following function: 

```
getCurrentRequest(): HearingModel {
        this.checkForExistingHearing();
        return this.modelHearing;
    }

```

Would it make sense to change this to the following and use that?

```
getCurrentRequest(): HearingModel {
        if(!this.modelHearing) {
                this.checkForExistingHearing();
        }
        return this.modelHearing;
    }

```

For reference checkForExistingHearing looks like this

```
    private checkForExistingHearing() {
        const localRequest = sessionStorage.getItem(this.newRequestKey);
        if (localRequest === null) {
            this.modelHearing = new HearingModel();
        } else {
            this.modelHearing = JSON.parse(localRequest);
        }
    }
```

AdminWebsite/AdminWebsite/ClientApp/src/app/common/model/participant.model.ts
mulvaneyjHMCTS: Change implemented

AdminWebsite/AdminWebsite/ClientApp/src/app/booking/participant/list/participant-list.component.ts
mulvaneyjHMCTS: This is probably as simplified as it can be

phillipjchambers:
```
        if (!this.canEdit || this.videoHearingsService.isConferenceClosed()) {
            return false;
        }

        if (!this.videoHearingsService.isHearingAboutToStart() && !particpant.addedDuringHearing) {
            return false;
        }

        return true;
```

phillipjchambers: or

```
        if (!this.canEdit || this.videoHearingsService.isConferenceClosed()) {
            return false;
        }

        return this.videoHearingsService.isHearingAboutToStart() || particpant.addedDuringHearing;
```

AdminWebsite/AdminWebsite/ClientApp/src/app/services/video-hearings.service.ts
phillipjchambers: Yeah, I think that's a good suggestion - getting it from mem each time if it's there instead of from local storage. I'd also consider doing that when this service is initialised perhaps, and then just always use `this.modelHearing`?

AdminWebsite/AdminWebsite/ClientApp/src/app/booking/participant/list/participant-list.component.ts
mulvaneyjHMCTS: Yeah I like the first option, implemented

phillipjchambers: Does the second not work?

AdminWebsite/AdminWebsite/ClientApp/src/app/services/video-hearings.service.ts
```diff
@@ -441,4 +445,18 @@ export class VideoHearingsService {
     getJudge(): ParticipantModel {
         return this.modelHearing.participants.find(x => x.is_judge);
     }
+
+    isConferenceClosed(): boolean {
+        return this.modelHearing.status === BookingStatus.Created && this.modelHearing.telephone_conference_id === '';
+    }
+
+    isHearingAboutToStart(): boolean {
+        if(this.modelHearing.scheduled_date_time && this.modelHearing.status) {
```
FarronF, line 454: this.modelHearing.status will only be populated if it is not a new hearing(so when editing). Based on Lorenzo's comments elsewhere this should be changed to only be allowed if the booking is confirmed which is not currently available on HearingModel.

phillipjchambers (dismissed):

AdminWebsite/AdminWebsite/ClientApp/src/app/services/video-hearings.service.ts
```diff
@@ -451,7 +452,7 @@ export class VideoHearingsService {
     }
 
     isHearingAboutToStart(): boolean {
-        if(this.modelHearing.scheduled_date_time && this.modelHearing.status) {
+        if (this.modelHearing.scheduled_date_time && this.modelHearing.status && this.modelHearing.isConfirmed) {
```
FarronF, line 455: I think we can remove this.modelHearing.status from this now, this was just to ensure that when creating a new booking within 30 minutes you would get locked out of most pages.

I think this.modelHearing.isConfirmed does this and more ensuring it's only confirmed bookings.

mulvaneyjHMCTS: You are correct. Implemented and pushed

phillipjchambers (approved):

PR #714, opened 2021-05-17
VIH-7012 multi date selection

phillipjchambers:

VIH-7012

### Change description ###

* allow users to select multiple, non-consecutive hearing dates

```
[x] Yes
[ ] No
```


AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
```diff
@@ -110,10 +110,25 @@ public async Task<ActionResult<HearingDetailsResponse>> Post([FromBody] BookHear
 
                 if (request.IsMultiDay)
                 {
+
```
shaed-parkar, line 113: To minimise how much this method grows, can you wrap the logic inside the isMultiDay block to its own method since the enclosing method doesn't rely on any of the variables produced

shaed-parkar (commented):
Looks good, just need to abstract some of the logic out of the route into its own method. Also  could you confirm the MDH hearing emails still work with the correct number of days in the email

shaed-parkar (approved):
LGTM

PR #731, opened 2021-06-02
VIH-7786 undefined value for title, firstname and lastname

alberto-hortiguela:
https://tools.hmcts.net/jira/browse/VIH-7786

### Change description ###
Title null value fix for empty salutation/title on Persons record

```
[ ] Yes
[x] No
```


AdminWebsite/AdminWebsite/ClientApp/src/app/common/model/participant-details.model.ts
```diff
@@ -21,9 +21,9 @@ export class ParticipantDetailsModel {
         isInterpretee: boolean
     ) {
         this.ParticipantId = participantId;
-        this.FirstName = firstName;
-        this.LastName = lastName;
-        this.Title = title === undefined ? '' : title;
+        this.FirstName = firstName === null ? '' : firstName;
```
robertbennett1998, line 24: Maybe change this to
`!firstName ? '' : firstName`
Then it will handle all cases?

alberto-hortiguela: Both yes, undefined was not handling the null
As a result an ugly "null" was displayed on the UI

robertbennett1998 (approved):
LGTM

PR #734, opened 2021-06-07
VIH-7664 Add support to send new  confirmation email to telephone Participant

PrasannaDommalapati:

https://tools.hmcts.net/jira/browse/VIH-7664

### Change description ###

- Added Mappi9ng for single and multi day confirmation email

```
[ ] Yes
[x] No
```


AdminWebsite/AdminWebsite/Models/EditHearingRequest.cs
```diff
@@ -44,6 +44,8 @@ public EditHearingRequest()
         /// </summary>
         public List<EditParticipantRequest> Participants { get; set; }
 
+        public List<EditTelephoneParticipantRequest> TelephoneParticipants { get; set; }
+
```
krishnapottekkat-hmcts, line 48: Do we need this new EditTelephoneParticipantRequest object altogether to append existing EditParticipantRequest object with a new ParticipantType ( to indicate it is a Video/Telephone participant) ?

PrasannaDommalapati: yes, Telephone participants are holding in a different model because it has fewer properties. When I was using the participant some of the fields that are not there in telephone participant are mandatory in the participant request.

simonwkelso (approved):
Looks good to me.

PR #736, opened 2021-06-14
VIH-7760 - changes done to fix retry of failed confirmation booking

khizarkhanu-hmcts:
https://tools.hmcts.net/jira/browse/VIH-7760 

### Change description ###
- changed AdminApi - UpdateBookingStatus method in Hearing API to handle excetions (normal, video, booking) for creation/cancellation actions.
- changed text in popup frames as per story/wire frame in sotry VIH-7560 

```
[ ] Yes
[ x] No
```

### DIFF
--- a/AdminWebsite/AdminWebsite.UnitTests/Controllers/HearingsController/UpdateBookingStatusTests.cs
+++ b/AdminWebsite/AdminWebsite.UnitTests/Controllers/HearingsController/UpdateBookingStatusTests.cs
@@ -458,7 +458,7 @@ namespace AdminWebsite.UnitTests.Controllers.HearingsController
         }
 
         [Test]
-        public async Task UpdateBookingStatus_returns_failed_status_response_when_there_is_exception()
+        public async Task UpdateBookingStatus_for_create_returns_failed_booking_status_in_response_when_there_is_nornal_exception()
         {
             _mocker.Mock<IUserIdentity>().Setup(x => x.GetUserIdentityName()).Returns("test");
             var updateCreatedStatus = new UpdateBookingStatusRequest
@@ -531,7 +531,7 @@ namespace AdminWebsite.UnitTests.Controllers.HearingsController
         }
 
         [Test]
-        public async Task UpdateBookingStatus_with_cancellation_returns_bad_request_when_throws_bookings_api_exception_notfound()
+        public async Task UpdateBookingStatus_with_cancellation_returns_not_found_when_throws_bookings_api_exception_of_status_code_not_found()
         {
             _mocker.Mock<IUserIdentity>().Setup(x => x.GetUserIdentityName()).Returns("test");
             var request = new UpdateBookingStatusRequest
--- a/AdminWebsite/AdminWebsite.UnitTests/Controllers/HearingsController/UpdateBookingStatusTests.cs
+++ b/AdminWebsite/AdminWebsite.UnitTests/Controllers/HearingsController/UpdateBookingStatusTests.cs
@@ -506,6 +506,28 @@ namespace AdminWebsite.UnitTests.Controllers.HearingsController
             
         }
 
+        public async Task UpdateBookingStatus_with_cancellation_returns_bad_request_when_throws_exception()
+        {
+            _mocker.Mock<IUserIdentity>().Setup(x => x.GetUserIdentityName()).Returns("test");
+            var request = new UpdateBookingStatusRequest
+            {
+                UpdatedBy = "test",
+                CancelReason = "",
+                Status = UpdateBookingStatus.Cancelled
+            };
+
+            var hearingId = Guid.NewGuid();
+
+            _mocker.Mock<IBookingsApiClient>().Setup(x => x.UpdateBookingStatusAsync(hearingId, request))
+                .ThrowsAsync(new Exception("Booking cancellation test exception."));
+
+            var response = await _controller.UpdateBookingStatus(hearingId, request);
+
+            var result = (BadRequestObjectResult)response;
+            result.StatusCode.Should().Be(StatusCodes.Status400BadRequest);
+
+        }
+
         private HearingDetailsResponse InitBookingForResponse(Guid hearingId)
         {
             var hearing = HearingResponseBuilder.Build()
--- a/AdminWebsite/AdminWebsite.UnitTests/Controllers/HearingsController/UpdateBookingStatusTests.cs
+++ b/AdminWebsite/AdminWebsite.UnitTests/Controllers/HearingsController/UpdateBookingStatusTests.cs
@@ -458,7 +458,59 @@ namespace AdminWebsite.UnitTests.Controllers.HearingsController
         }
 
         [Test]
-        public void UpdateBookingStatus_throws_when_bookings_api_exception()
+        public async Task UpdateBookingStatus_returns_failed_status_response_when_there_is_exception()
+        {
+            _mocker.Mock<IUserIdentity>().Setup(x => x.GetUserIdentityName()).Returns("test");
+            var updateCreatedStatus = new UpdateBookingStatusRequest
+            {
+                UpdatedBy = "test",
+                CancelReason = "",
+                Status = UpdateBookingStatus.Created
+            };
+
+            var hearingId = Guid.NewGuid();
+            var expectedConferenceDetailsResponse = new ConferenceDetailsResponse
+            {
+                Id = Guid.NewGuid(),
+                HearingId = hearingId,
+                MeetingRoom = new MeetingRoomResponse
+                {
+                    AdminUri = "admin",
+                    JudgeUri = "judge",
+                    ParticipantUri = "participant"
+                }
+            };
+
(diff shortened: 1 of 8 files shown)


AdminWebsite/AdminWebsite/ClientApp/src/app/popups/confirm-booking-failed-popup/confirm-booking-failed-popup.component.spec.ts
```diff
@@ -23,17 +24,26 @@ describe('SaveFailedPopupComponent', () => {
...
 
-    it('should emit event when the ok button is clicked', () => {
+    it('should emit event when the close button is clicked', () => {
         spyOn(component.close, 'emit');
         buttonTryAgain.nativeElement.click();
         expect(component.close.emit).toHaveBeenCalled();
     });
+
+    it('should have heading Booking confirmation failed!', () => {
+        expect(headingText.nativeElement.innerText).toEqual('Booking confirmation failed!');
```
omio-ma, line 43: I don't think we should be testing text content in unit test, can we instead check if the element with an id/data-handler exist instead?

AdminWebsite/AdminWebsite/ClientApp/src/app/popups/wait-popup/wait-popup.component.spec.ts
```diff
@@ -16,10 +19,16 @@ describe('WaitPopupComponent', () => {
...
+        headingText = de.query(By.css('.govuk-heading-m'));
         fixture.detectChanges();
     });
 
     it('should create wait pop up component', () => {
         expect(component).toBeTruthy();
     });
+
+    it('should have heading Confirming bookings can take up to 60 seconds to complete.', () => {
+        expect(headingText.nativeElement.innerText).toEqual('Confirming bookings can take up to 60 seconds to complete.');
```
omio-ma, line 32: As above, I strongly believe we dont need to test the text content, I would rather not have this test at all or add an id or data handler to the element and assert that element exists

AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
```diff
@@ -455,30 +451,60 @@ public async Task<ActionResult> GetHearingById(Guid hearingId)
...
+                        new UpdateBookingStatusRequest
+                        {
+                            Status = BookingsApi.Contract.Requests.Enums.UpdateBookingStatus.Failed,
+                            UpdatedBy = "System",
+                            CancelReason = string.Empty
+                        });
+
+                    return Ok(new UpdateBookingStatusResponse { Success = false, Message = errorMessage });
+                }
+                if (ex is BookingsApiException)
```
omio-ma, line 500: can we chain this exception instead of adding an if logic here: 
```
catch (BookingsApiException e)
{
//implementation
}
 catch (Exception ex)
{
//implementation
}
```

AdminWebsite/AdminWebsite/ClientApp/src/app/popups/confirm-booking-failed-popup/confirm-booking-failed-popup.component.spec.ts
PrasannaDommalapati: it is tested because the previous implementation had different text and bug was raised. we made sure the text is correct.

AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
PrasannaDommalapati: you will have duplicated lines of code the booking service for the bad request of we do like this.

khizarkhanu-hmcts: I had thought thur this and I think we will be duplicating code in each catch, the code structure was previously not that good, so to avoid refactoring we have to live with this.

AdminWebsite/AdminWebsite.UnitTests/Controllers/HearingsController/UpdateBookingStatusTests.cs
```diff
@@ -458,7 +458,57 @@ public async Task UpdateBookingStatus_returns_false_response_when_conference_exi
         }
 
         [Test]
-        public void UpdateBookingStatus_throws_when_bookings_api_exception()
+        public async Task UpdateBookingStatus_returns_failed_status_response_when_there_is_exception()
```
robertbennett1998, line 461: Maybe change the name as it sounds like it should not return a success HTTP status code. Not just set the status flag.

AdminWebsite/AdminWebsite.UnitTests/Controllers/HearingsController/UpdateBookingStatusTests.cs
```diff
@@ -473,7 +523,79 @@ public void UpdateBookingStatus_throws_when_bookings_api_exception()
...
-            Assert.ThrowsAsync<BookingsApiException>(() => _controller.UpdateBookingStatus(hearingId, request));
+            var response = await _controller.UpdateBookingStatus(hearingId, request);
+
+            var result = (BadRequestObjectResult)response;
+            result.StatusCode.Should().Be(StatusCodes.Status400BadRequest);
+            
+        }
+
+        [Test]
+        public async Task UpdateBookingStatus_with_cancellation_returns_bad_request_when_throws_bookings_api_exception_notfound()
```
robertbennett1998, line 534: rename to returns not found instead of bad request

AdminWebsite/AdminWebsite/ClientApp/src/app/popups/confirm-booking-failed-popup/confirm-booking-failed-popup.component.html
```diff
@@ -1,19 +1,11 @@
...
-      </div>
-    </div>
-
+    <h1 class="govuk-heading-m vh-ml70 vh-mr70">Booking confirmation failed!</h1>
     <div class="content">
+      <span id="content-text">Your booking cannot be confirmed. Please try again later.</span>
       <div class="govuk-grid-row">
         <div class="vh-top">
-          <button id="btnTryAgain" class="govuk-button vh-mr25" data-module="govuk-button" (click)="ok()" alt="Ok">Ok</button>
+          <button id="btnTryAgain" class="govuk-button vh-mr25" data-module="govuk-button" (click)="closePopup()" alt="Close">Close</button>
```
robertbennett1998, line 8: Should probably be using the translate service for the button text as we want it to translate to Welsh

AdminWebsite/AdminWebsite/ClientApp/src/app/popups/wait-popup/wait-popup.component.html
```diff
@@ -1,7 +1,6 @@
 <div id="waitPopup" class="vh-popup-overlay">
   <div class="vh-popup vh-popup-xsmall">
-    <h1 class="govuk-heading-m vh-ml70 vh-mr70">Please wait while we process your request</h1>
-
+    <h1 class="govuk-heading-m vh-ml70 vh-mr70">Confirming bookings can take up to 60 seconds to complete.</h1>
```
robertbennett1998, line 3: Again, I think this should be a translated string. I know you haven't put this in but it will likely be raised as a bug when the Welsh is tested

robertbennett1998: This applies to all strings displayed to the user.

AdminWebsite/AdminWebsite/ClientApp/src/app/popups/confirm-booking-failed-popup/confirm-booking-failed-popup.component.html
khizarkhanu-hmcts: as dicussed, Admin does not have language support

AdminWebsite/AdminWebsite/ClientApp/src/app/popups/wait-popup/wait-popup.component.html
khizarkhanu-hmcts: as dicussed, Admin does not have language support

AdminWebsite/AdminWebsite.UnitTests/Controllers/HearingsController/UpdateBookingStatusTests.cs
khizarkhanu-hmcts: fixed title of test case to include not found

AdminWebsite/AdminWebsite.UnitTests/Controllers/HearingsController/UpdateBookingStatusTests.cs
khizarkhanu-hmcts: fixed test case title to include when you create a booking with exception, the result is back is ok but with failed booking status. (functionality requirement)

robertbennett1998 (approved):
LGTM

PR #740, opened 2021-06-21
VIH-7760 - changes done to fix retry of failed confirmation booking

khizarkhanu-hmcts:

https://tools.hmcts.net/jira/browse/VIH-7760

### Change description ###

changes made to fix timing issue in higher environments, if comfirmation fails, it should not raise a reload event rather set the status in local storage.

```
[ ] Yes
[x ] No
```

### DIFF
--- a/AdminWebsite/AdminWebsite/ClientApp/src/app/bookings-list/booking-details/booking-details.component.ts
+++ b/AdminWebsite/AdminWebsite/ClientApp/src/app/bookings-list/booking-details/booking-details.component.ts
@@ -223,7 +223,9 @@ export class BookingDetailsComponent implements OnInit, OnDestroy {
         if (status === UpdateBookingStatus.Cancelled) {
             this.showCancelBooking = false;
         }
+        this.hearing.Status = UpdateBookingStatus[status];
         this.persistStatus(status);
+        if (status === UpdateBookingStatus.Failed) return;
         this.$subscriptions.push(
             this.videoHearingService.getHearingById(this.hearingId).subscribe(
                 newData => {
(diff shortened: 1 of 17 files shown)


AdminWebsite/AdminWebsite/ClientApp/src/app/bookings-list/booking-details/booking-details.component.ts
```diff
@@ -223,7 +223,11 @@ export class BookingDetailsComponent implements OnInit, OnDestroy {
         if (status === UpdateBookingStatus.Cancelled) {
             this.showCancelBooking = false;
         }
+        this.hearing.Status = UpdateBookingStatus[status];
```
robertbennett1998, line 226: Can this not just be:
`this.hearing.Status = status`
as status is already the correct type.

AdminWebsite/AdminWebsite/ClientApp/src/app/bookings-list/booking-details/booking-details.component.ts
```diff
@@ -223,7 +223,11 @@ export class BookingDetailsComponent implements OnInit, OnDestroy {
         if (status === UpdateBookingStatus.Cancelled) {
```
robertbennett1998, line 223: Could you also add a unit test to make sure `this.videoHearingService.getHearingById(this.hearingId)` is never called if the status is failed.

robertbennett1998 (changes requested):
Overall this is fine and will fix the issue but it seems like there is a bigger underlying cause. This issue shouldn't exist anyways and suggests that there may need to be some refactoring.

AdminWebsite/AdminWebsite/ClientApp/src/app/bookings-list/booking-details/booking-details.component.ts
khizarkhanu-hmcts: I have now fixed to handle when status is failed so other unit test are not impacted.

AdminWebsite/AdminWebsite/ClientApp/src/app/bookings-list/booking-details/booking-details.component.ts
khizarkhanu-hmcts: it is already there and this is not called, pls refer to test 'should show pop up if the confirm failed' in booking-details.component.spec.ts

robertbennett1998 (approved):
LGTM! Previous comment still applies.

PR #755, opened 2021-07-06
VIH-7932 Judge accounts available for selection for individual participants

robertbennett1998:
https://tools.hmcts.net/jira/browse/VIH-7932

### Change description ###
- Added temporary filtering in PostPersonBySearchTerm to hide hearings reform judge accounts and ejud accounts.

```
[ ] Yes
[x] No
```

### DIFF
--- a/AdminWebsite/AdminWebsite.UnitTests/Controllers/PersonsControllerTest.cs
+++ b/AdminWebsite/AdminWebsite.UnitTests/Controllers/PersonsControllerTest.cs
@@ -72,19 +72,19 @@ namespace AdminWebsite.UnitTests.Controllers
             };
             _response.Add(additionalParticipantToReturn);
             
+            var searchTerm = "ado";
+            
             var expectedResponse = new List<PersonResponse>();
             expectedResponse.Add(_response[0]);
             expectedResponse.Add(additionalParticipantToReturn);
             
-            _bookingsApiClient.Setup(x => x.PostPersonBySearchTermAsync(It.IsAny<SearchTermRequest>()))
+            _bookingsApiClient.Setup(x => x.PostPersonBySearchTermAsync(It.Is<SearchTermRequest>(searchTermRequest => searchTermRequest.Term == searchTerm)))
                 .ReturnsAsync(_response);
 
             _userAccountService.Setup(x => x.GetJudgeUsers()).ReturnsAsync(new List<JudgeResponse>());
 
-            _bookingsApiClient.Setup(x => x.PostJudiciaryPersonBySearchTermAsync(It.IsAny<SearchTermRequest>()))
+            _bookingsApiClient.Setup(x => x.PostJudiciaryPersonBySearchTermAsync(It.Is<SearchTermRequest>(searchTermRequest => searchTermRequest.Term == searchTerm)))
                 .ReturnsAsync(new List<PersonResponse>());
-            
-            var searchTerm = "ado";
 
             // Act
             var result = await _controller.PostPersonBySearchTerm(searchTerm);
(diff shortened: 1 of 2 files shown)


AdminWebsite/AdminWebsite/Controllers/PersonsController.cs
```diff
@@ -58,7 +58,17 @@ public async Task<ActionResult<IList<PersonResponse>>> PostPersonBySearchTerm([F
                 term = _encoder.Encode(term);
                 var searchTerm = new SearchTermRequest(term);
 
-                var personsResponse = await _bookingsApiClient.PostPersonBySearchTermAsync(searchTerm);
+                var excludedEmails = (await _userAccountService.GetJudgeUsers()).Select(x => x.Email.ToLower()).ToList();
```
omio-ma, line 61: A suggestion here, can we use toLower to toLowerInvariant.
I will be honest, I had to refer back to what the difference is, basically it ignore culture specefic characters and will use lowercase in english I believe no matter the case. 

Here is a stack overflow post I found about it: https://stackoverflow.com/questions/6225808/string-tolower-and-string-tolowerinvariant

AdminWebsite/AdminWebsite/Controllers/PersonsController.cs
```diff
@@ -58,7 +58,17 @@ public async Task<ActionResult<IList<PersonResponse>>> PostPersonBySearchTerm([F
...
+                var persons = await _bookingsApiClient.PostJudiciaryPersonBySearchTermAsync(searchTerm);
+                var excludedUsernames = persons.Where(x => x.Username != null).Select(x => x.Username.ToLower()).ToList();
+                var excludedContactEmails = persons.Where(x => x.ContactEmail != null).Select(x => x.ContactEmail.ToLower()).ToList();
```
omio-ma, line 64: can line 63 and 64 be done in one query? Have we checked to see if using an OR check in the where clause produces the same result?

AdminWebsite/AdminWebsite.UnitTests/Controllers/PersonsControllerTest.cs
```diff
@@ -55,35 +56,183 @@ public void Setup()
...
+            
+            var searchTerm = "ado";
+
+            // Act
+            var result = await _controller.PostPersonBySearchTerm(searchTerm);
+
+            // Assert
+            var okObjectResult = result.Result.Should().BeAssignableTo<OkObjectResult>().Which;
+            okObjectResult.Value.Should().BeEquivalentTo(expectedResponse);
+            _bookingsApiClient.Verify(x => x.PostPersonBySearchTermAsync(It.Is<SearchTermRequest>(request => request.Term == searchTerm)), Times.Once);
```
omio-ma, line 95: I was able to do this in nsubstitute and I think we can do this here too, it is however just a suggestion to consider. 

Can we add this query in line 79, so it would mean that we only return the the expected response for the given argument. Ultimately based on the response you get from that mocked client calls you can get the expected response so you don't have to do this added assertions. We can then do the same for the bookingsApiClient mock setup.

AdminWebsite/AdminWebsite.UnitTests/Controllers/PersonsControllerTest.cs
```diff
@@ -55,35 +56,183 @@ public void Setup()
...
+                MiddleNames = "No",
+                Username = "jackfilter@hmcts.net"
+            };
+            _response.Add(additionalParticipantToReturn);
+            
+            var expectedResponse = new List<PersonResponse>();
+            expectedResponse.Add(_response[0]);
+            expectedResponse.Add(additionalParticipantToReturn);
+            
+            _bookingsApiClient.Setup(x => x.PostPersonBySearchTermAsync(It.IsAny<SearchTermRequest>()))
```
omio-ma, line 79: following on my comment below, so here instead of using It.IsAny<SearchTermRequest>() we use It.Is<SearchTermRequest>(request => request.Term == searchTerm) and then return the response

kevin-hmcts (approved):
LGTM

kevin-hmcts (approved):

PR #757, opened 2021-07-12
VIH-7965 Judge details disappearing from booking form after save

robertbennett1998:
https://tools.hmcts.net/jira/browse/VIH-7965

### Change description ###

- Refactored process new participant method to remove the side effect of modifying the new participants list. 
- Refactored process new participant method to allow addition of a participant that will be removed.

```
[ ] Yes
[x] No
```

### DIFF
--- a/AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
+++ b/AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
@@ -269,12 +269,21 @@ namespace AdminWebsite.Controllers
                 var removedParticipantIds = originalHearing.Participants.Where(p => request.Participants.All(rp => rp.Id != p.Id)).Select(x => x.Id).ToList();
 
                 foreach (var participant in request.Participants)
+                {
                     if (!participant.Id.HasValue)
-                        await _hearingsService.ProcessNewParticipants(hearingId, participant, originalHearing,
-                            usernameAdIdDict, newParticipants);
+                    {
+                        if (await _hearingsService.ProcessNewParticipant(hearingId, participant,
+                            removedParticipantIds,
+                            originalHearing,
+                            usernameAdIdDict) is { } newParticipant)
+                        {
+                            newParticipants.Add(newParticipant);
+                        }
+                    }
                     else
                     {
-                        var existingParticipant = originalHearing.Participants.FirstOrDefault(p => p.Id.Equals(participant.Id));
+                        var existingParticipant =
+                            originalHearing.Participants.FirstOrDefault(p => p.Id.Equals(participant.Id));
                         if (existingParticipant == null || string.IsNullOrEmpty(existingParticipant.UserRoleName))
                         {
                             continue;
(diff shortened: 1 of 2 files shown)


AdminWebsite/AdminWebsite/ClientApp/src/app/services/bookings-persist.service.ts
```diff
@@ -35,6 +35,7 @@ export class BookingPersistService {
                 hearingUpdate.HearingCaseNumber = hearing.cases && hearing.cases.length > 0 ? hearing.cases[0].number : '';
                 hearingUpdate.StartTime = newStartDate;
                 hearingUpdate.Duration = hearing.scheduled_duration;
+                hearingUpdate.CourtRoomAccount = hearing.participants.find(x => x.is_judge)?.username;
```
omio-ma, line 38: Are we trying to identify whether or not the participant is a judge here? 
If that is the case, I think we might want to use is_courtroom_account as the is_judge flag is evaluated based on the hearing role name which has been mentioned to be not reliable. However I might be wrong we would need to check whether you get the same result with is_courtroom_account  (which should be available in the ParticipantModel)

robertbennett1998: Yeah. I can change it, it is available, what does it actually mean, is_courtroom_account? Does this include panel members as it needs to literally only be the Judge. Due to where it is in the code I would hope that the HearingRole is set correctly.

robertbennett1998: By the looks of it using is_courtroom_account would prevent this for working for eJud?
`
is_courtroom_account: judge.account_type === JudgeAccountType.Courtroom
`

`
export enum JudgeAccountType {
    Courtroom = 'Courtroom',
    Judiciary = 'Judiciary'
}
`

robertbennett1998: This code is also used within the video-hearings.service.ts to get the judge:
`   
getJudge(): ParticipantModel {
        return this.modelHearing.participants.find(x => x.is_judge);
}
`

omio-ma: Do you mean for eJud we are not getting any value back for the account_type? 

The reason why I have raised this is because is_judge appears to be evaluated by the court room name and hearing role name which also appears to be the rule that we use to identify whether the participant is a judge or not in the BookNewHearing endpoint (of the admin web). I also assumed that was the correct property to use to check whether a participant is a judge or not however in the standup I think it was mentioned that we shouldn't rely on the hearing role name, and to identify the judge we could use the account type which every judge (including ejud) should have either set to Courtroom or Judiciary.

omio-ma: But thanks for trying it out, if your solution works with is_judge I would say go ahead with it, I just had a quick look I can see that we are using is_judge in our client app to evaluate a judge

omio-ma (approved):

PR #763, opened 2021-07-16
VIH-7902 unique vh joining link on hearing confirmation

alberto-hortiguela:
https://tools.hmcts.net/jira/browse/VIH-7902

### Change description ###
Removed quickJoin from URL

```
[ ] Yes
[x] No
```

### DIFF
--- a/AdminWebsite/AdminWebsite/ClientApp/src/app/shared/page-url.constants.ts
+++ b/AdminWebsite/AdminWebsite/ClientApp/src/app/shared/page-url.constants.ts
@@ -17,5 +17,5 @@ export let PageUrls = {
     EditParticipantSearch: '/edit-participant-search',
     EditParticipant: '/edit-participant',
     Login: '/login',
-    JoinLink: '/quickjoin',
+    JoinLink: '/quickjoin'
 };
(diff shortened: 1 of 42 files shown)


AdminWebsite/AdminWebsite/ClientApp/src/app/shared/page-url.constants.ts
```diff
@@ -16,6 +16,5 @@ export let PageUrls = {
     DeleteParticipant: '/delete-participant',
     EditParticipantSearch: '/edit-participant-search',
     EditParticipant: '/edit-participant',
-    Login: '/login',
-    JoinLink: '/quickjoin'
+    Login: '/login'
```
omio-ma, line 19: just double checking this wasnt removed by mistake? Dont we need the "quickjoin" link for the new endpoint

alberto-hortiguela: they needed originally for the scan readers but BAs have confirmed it is not need at all as we copy a link to the clipboard directly.
This is the reason for this last change

omio-ma (approved):

PR #813, opened 2021-09-16
VIH-8196 Hide Adding staff member with feature flags

PrasannaDommalapati:
https://tools.hmcts.net/jira/browse/VIH-8196

### Change description ###

Hide Adding staff member with feature flags

```
[ ] Yes
[x] No
```


AdminWebsite/AdminWebsite/ClientApp/src/app/booking/assign-judge/assign-judge.component.ts
```diff
@@ -87,6 +90,7 @@ export class AssignJudgeComponent extends BookingBaseComponent implements OnInit
 
     ngOnInit() {
         this.failedSubmission = false;
+        this.featureService.getFeatureToggles().subscribe(t => (this.showStaffMemberFeature = t.staff_member));
```
omio-ma, line 93: needs to be unsubscribed when this component gets destroyed, or if we only need to call this once we can use the first or take operators from rxjs

ZJianWang (approved):
LGTM

AdminWebsite/AdminWebsite/ClientApp/src/app/services/feature-toggle.service.spec.ts
```diff
@@ -0,0 +1,42 @@
...
+    });
+
+    it('should be created', () => {
+        expect(service).toBeTruthy();
+    });
+    it('should return the feature config for staff member feature turned on', () => {
+        featureConfig.staff_member = true;
+        clientApiSpy.getFeatureToggles.and.returnValue(of(featureConfig));
+
+        service.getFeatureToggles().subscribe(config => {
```
omio-ma, line 30: this is interesting as we are subscribing to an observable in a test so we might have to unsubscribe from here as well, I would suggest however either to use async function here and await or use fakeasync instead.

AdminWebsite/AdminWebsite.UnitTests/Controllers/FeatureControllerTests.cs
```diff
@@ -0,0 +1,30 @@
...
+            _mocker = AutoMock.GetLoose();
+            _mocker.Mock<IOptions<FeatureToggleConfiguration>>().Setup(opt => opt.Value).Returns(new FeatureToggleConfiguration()
+            {
+                StaffMember = true
+            });
+
+            var _controller = _mocker.Create<FeatureController>();
+            var result = _controller.GetFeatureToggles();
+
+            result.Value.StaffMember.Should().Be(true);
```
omio-ma, line 27: you can use .Should().BeTrue() here as the value here will be boolean

ZJianWang (approved):

kevin-hmcts (approved):
LGTM

PR #868, opened 2022-02-25
VIH-8812 Fix for interpreter incorrectly showing in hearing role selection when valid interpretee (not appraiser/observer) exists already

will-craig:
https://tools.hmcts.net/jira/browse/VIH-8812

### Change description ###
Moved notObserver logic up to base class and now utilising this same logic for populating hearing role list, so that an interpreter can not be added if an observer/appraiser exists only. Also added supporting unit test.

```
[ ] Yes
[ x] No
```

### DIFF
--- a/AdminWebsite/AdminWebsite/ClientApp/src/app/booking/add-participant-base/add-participant-base.component.ts
+++ b/AdminWebsite/AdminWebsite/ClientApp/src/app/booking/add-participant-base/add-participant-base.component.ts
@@ -324,7 +324,7 @@ export abstract class AddParticipantBaseDirective extends BookingBaseComponent i
 
     private hearingHasInterpretees(): boolean {
         return this.hearing.participants.some(
-            p => p.user_role_name === 'Individual' && p.hearing_role_name !== Constants.HearingRoles.Interpreter && this.isNotAnObserver(p)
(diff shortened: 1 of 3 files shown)


AdminWebsite/AdminWebsite/ClientApp/src/app/booking/add-participant-base/add-participant-base.component.ts
```diff
@@ -398,4 +397,10 @@ export abstract class AddParticipantBaseDirective extends BookingBaseComponent i
             this.isInterpreter = false;
         }
     }
+
+    protected isNotAnObserver(participant): boolean {
```
oliver-scott, line 401: Just wondering if it would be better to reverse the logic and make this `isAnObserver`. It might make it simpler to re-use as you could call it with `this.isAnObserver() `rather than `!this.isNotAnObserver()`.

will-craig: Sure, that makes sense. Will update it.

oliver-scott (approved):

AWildBugAppeared (approved):

PR #952, opened 2022-08-30
VIH-9280 Updated the regular expression used on first and last name input on add participants

will-craig:
https://tools.hmcts.net/jira/browse/VIH-9280

### Change description ###
Updated the regular expression used on first and last name input on add-participant-base.component.ts to make it valid for transformation into a valid email address on the user api. Including no non alpha numerical charecters trailing or starting a word and no double symbols i.e. will..iam 

```
[ ] Yes
[X] No
```


oliver-scott:
it might be worth mirroring this validation in the backend API (which I think would be bookings API).

will-craig:
> it might be worth mirroring this validation in the backend API (which I think would be bookings API).
Sure, can do

marcogagliardi (dismissed):
👌🏽

oliver-scott (approved):

PR #963, opened 2022-09-20
VIH-9309 - moved get status out of the booking hearing method and cre…

marcogagliardi:
https://tools.hmcts.net/jira/browse/VIH-9309

### Change description ###
moved get status out of the booking hearing method and created a standalone call for it. Created timer in the client to get status


AdminWebsite/AdminWebsite/ClientApp/src/app/booking/summary/summary.component.ts
```diff
@@ -309,6 +319,20 @@ export class SummaryComponent implements OnInit, OnDestroy {
         }
     }
 
+    async oberserableTimer(hearingId: string): Promise<UpdateBookingStatusResponse> {
```
oliver-scott, line 322: could we name this function something like `getHearingStatus`?

AdminWebsite/AdminWebsite/ClientApp/src/app/booking/summary/summary.component.ts
```diff
@@ -309,6 +319,20 @@ export class SummaryComponent implements OnInit, OnDestroy {
...
+        let hearingStatusResponse;
+
+        const source = timer(0, 5000);
+        const schedule = source.subscribe(async counter => {
+            hearingStatusResponse = await this.hearingService.getStatus(hearingId);
+            if (hearingStatusResponse.success || counter === 6) {
```
oliver-scott, line 328: what is the 6 for here - is that the timeout?

if so we'll need to poll for longer than 30 seconds, as large bookings can take longer.

maybe go for 50 seconds, as per the ticket

AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
```diff
@@ -92,24 +92,25 @@ public async Task<ActionResult<HearingDetailsResponse>> Post([FromBody] BookHear
...
                     return Created("", hearingDetailsResponse);
                 }
-                var judgeExists = request.BookingDetails.Participants?.Any(x => x.HearingRoleName == RoleNames.Judge) ?? false;
-                if (judgeExists)
-                {
-                    await GetConferenceStatus(hearingDetailsResponse.Id, 
-                        $"Failed to get the conference from video api, possibly the conference was not created or the kinly meeting room is null - hearingId: {hearingDetailsResponse.Id}");
-                }
+
+                // var judgeExists = request.BookingDetails.Participants?.Any(x => x.HearingRoleName == RoleNames.Judge) ?? false;
```
oliver-scott, line 100: can we remove this commented out code?

AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
```diff
@@ -414,48 +434,129 @@ public async Task<ActionResult> GetHearingById(Guid hearingId)
...
+                }
+            }
+            catch (VideoApiException ex)
+            {
+                _logger.LogError(ex, "Failed to confirm a hearing. {ErrorMessage}", errorMessage);
+                _logger.LogError("There was an unknown error for hearing {Hearing}. Updating status to failed",
+                    hearingId);
+
+                // Set the booking status to failed as the video api failed
+                //await UpdateFailedBookingStatus(hearingId);
```
oliver-scott, line 520: can we remove this commented out code?

AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
```diff
@@ -414,48 +434,129 @@ public async Task<ActionResult> GetHearingById(Guid hearingId)
...
+                return Ok(new UpdateBookingStatusResponse {Success = false, Message = errorMessage});
+            }
+        }
+
+        /// <summary>
+        ///     Update the failed hearing status.
+        /// </summary>
+        /// <param name="hearingId">The hearing id</param>
+        /// <returns>Success status</returns>
+        [HttpGet("{hearingId}/update-status")]
```
oliver-scott, line 531: this should be a Put rather than a Get since it modifies data.

think it's probably also worth renaming it so that it's clear it only changes the status to failed.

something like `{hearingId}/failed-status`

`public async Task <IActionResult> UpdateFailedBookingStatus(Guid hearingId)`

AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
```diff
@@ -414,48 +434,129 @@ public async Task<ActionResult> GetHearingById(Guid hearingId)
...
+        public async Task<IActionResult> UpdateHearingStatus(Guid hearingId)
+        {
+            var errorMessage =
+                $"Failed to get the conference from video api, possibly the conference was not created or the kinly meeting room is null - hearingId: {hearingId}";
+            try
+            {
+                await UpdateFailedBookingStatus(hearingId);
+                return Ok(new UpdateBookingStatusResponse {Success = false, Message = errorMessage});
+            }
+            catch (Exception e)
```
oliver-scott, line 545: what happens if the hearing doesn't exist? Could we add a test to assert that it returns a 404?

AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
```diff
@@ -414,48 +434,129 @@ public async Task<ActionResult> GetHearingById(Guid hearingId)
...
+                        Success = true,
+                        TelephoneConferenceId = conferenceDetailsResponse.MeetingRoom.TelephoneConferenceId
+                    });
+                }
+                else
+                {
+                    return Ok(new UpdateBookingStatusResponse {Success = false, Message = errorMessage});
+                }
+            }
+            catch (VideoApiException ex)
```
oliver-scott, line 513: what happens if the hearing doesn't exist? Could we add a test to assert that it returns a 404/not found?

AdminWebsite/AdminWebsite/ClientApp/src/app/booking/summary/summary.component.ts
marcogagliardi: changed to 20 as any 5 seconds for 10 times is 50s

AdminWebsite/AdminWebsite/ClientApp/src/app/booking/summary/summary.component.ts
marcogagliardi: name changed

AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
marcogagliardi: comments removed

AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
marcogagliardi: comments removed

AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
marcogagliardi: api changed

PR #964, opened 2022-09-21
VIH-9202 Upgraded Target Framework to dotnet 6 

will-craig:
https://tools.hmcts.net/jira/browse/VIH-9202

### Change description ###
Upgraded target framework to .NET6

```
[ ] Yes
[X] No
```

### DIFF
--- a/AdminWebsite/AdminWebsite.AcceptanceTests/appsettings.json
+++ b/AdminWebsite/AdminWebsite.AcceptanceTests/appsettings.json
@@ -9,7 +9,8 @@
         "Authority": "https://login.microsoftonline.com/",
         "GraphApiBaseUri": "https://graph.microsoft.com/",
         "RedirectUri": " https://localhost:5400/login",
-        "PostLogoutRedirectUri": "https://localhost:5400/"
+        "PostLogoutRedirectUri": "https://localhost:5400/",
+        "ResourceId": "resourceID"
     },
     "IsLive": false,
     "UsingEjud": false,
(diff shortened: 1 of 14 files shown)


premk-hmcts (approved):

AdminWebsite/AdminWebsite.AcceptanceTests/appsettings.json
```diff
@@ -9,7 +9,8 @@
         "Authority": "https://login.microsoftonline.com/",
         "GraphApiBaseUri": "https://graph.microsoft.com/",
         "RedirectUri": " https://localhost:5400/login",
-        "PostLogoutRedirectUri": "https://localhost:5400/"
+        "PostLogoutRedirectUri": "https://localhost:5400/",
+        "ResourceId": "resourceID"
```
oliver-scott, line 13: don't think the `ResourceId` is needed here. It gets replaced by the client id in the frontend:

```
configService.getClientSettings().subscribe(clientSettings => {
            const resource = clientSettings.resource_id ? clientSettings.resource_id : `api://${clientSettings.client_id}`;
```

will-craig: You're right, it's not needed in the pipeline. It is needed if you want to run the admin web Acceptance Tests locally, else it throws an exception that the value is missing when when it tries to parse the file in the configuration hook. I figured I'd leave the change in so people dont need to add it in, in future.

oliver-scott: right ok. I already have the ResourceId in secrets, probably why it didn't fail for me. But yes, that sounds fine

PR #972, opened 2022-09-28
VIH-9309 - gateway timeout fix

marcogagliardi:
https://tools.hmcts.net/jira/browse/VIH-9309

moved get status to client


marcogagliardi:
no needed as it is a duplicate

PR #974, opened 2022-10-05
VIH-9315 Save working hours

AWildBugAppeared:
https://tools.hmcts.net/jira/browse/VIH-9315


AdminWebsite/AdminWebsite.UnitTests/Controllers/WorkHoursControllerTests.cs
```diff
@@ -0,0 +1,49 @@
...
+
+        private WorkHoursController _controller;
+
+        [SetUp]
+        public void Setup()
+        {
+            _failedUsernames = new List<string> { "failedusername@test.com" };
+
+            _bookingsApiClientMock = new Mock<IBookingsApiClient>();
+            _bookingsApiClientMock.Setup(x => x.SaveWorkHoursAsync(It.IsAny<List<UploadWorkHoursRequest>>()))
```
oliver-scott, line 28: could we maybe move this failed usernames setup into the test instead, so that it won't return failed usernames if we add more tests?

AdminWebsite/AdminWebsite/Controllers/WorkHoursController.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public WorkHoursController(IBookingsApiClient bookingsApiClient)
+        {
+            _bookingsApiClient = bookingsApiClient;
+        }
+
+        [HttpPost]
+        [AllowAnonymous]
+        [SwaggerOperation(OperationId = "UploadWorkHours")]
+        [ProducesResponseType(typeof(List<string>), (int)HttpStatusCode.OK)]
```
oliver-scott, line 27: rather than returning a list of strings, could we maybe return an `UploadWorkHoursResponse` object instead, with a `List<string> FailedUsernames` as a property?

it would make it easier to change later if we need to return additional information

AdminWebsite/AdminWebsite/Controllers/WorkHoursController.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        private readonly IBookingsApiClient _bookingsApiClient;
+
+        public WorkHoursController(IBookingsApiClient bookingsApiClient)
+        {
+            _bookingsApiClient = bookingsApiClient;
+        }
+
+        [HttpPost]
+        [AllowAnonymous]
```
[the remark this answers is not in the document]
AWildBugAppeared: Every single time

PR #975, opened 2022-10-06
VIH-9309 test coverage fixes and original change

will-craig:
(https://tools.hmcts.net/jira/browse/VIH-9309)

### Change description ###
Added client side polling, to check the booking confirmation status in the video api

```
[ ] Yes
[X] No
```


AdminWebsite/AdminWebsite/ClientApp/src/app/booking/summary/summary.component.ts
```diff
@@ -243,72 +240,83 @@ export class SummaryComponent implements OnInit, OnDestroy {
...
-                this.router.navigate([PageUrls.BookingConfirmation]);
             } catch (error) {
                 this.logger.error(`${this.loggerPrefix} Failed to save booking.`, error, { payload: this.hearing });
                 this.setError(error);
             }
         }
     }
 
+    async processBooking(hearingDetailsResponse, hearingStatusResponse): Promise<void> {
+        if (hearingStatusResponse?.success) {
```
oliver-scott, line 269: From the code coverage results it looks like we might be missing coverage for when this success response is true. Perhaps we are missing a test?

AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
```diff
@@ -82,34 +82,22 @@ public async Task<ActionResult<HearingDetailsResponse>> Post([FromBody] BookHear
...
-                if(hearingDetailsResponse.Status == BookingStatus.Failed)
-                {
-                    return Created("", hearingDetailsResponse);
-                }
-                var judgeExists = request.BookingDetails.Participants?.Any(x => x.HearingRoleName == RoleNames.Judge) ?? false;
-                if (judgeExists)
-                {
-                    await GetConferenceStatus(hearingDetailsResponse.Id, 
-                        $"Failed to get the conference from video api, possibly the conference was not created or the kinly meeting room is null - hearingId: {hearingDetailsResponse.Id}");
-                }
```
will-craig, line 104: @marcogagliardi Have just realised, this condition still needs to be on to decided whether we check the conference status, should've noticed it when I did my refactor -- when a booking is made without a judge, it isnt confirmed so we need to maybe add this condition to the client side. If booking doesnt contain a judge -- dont check the conference status

AdminWebsite/AdminWebsite/ClientApp/src/app/booking/summary/summary.component.ts
marcogagliardi: added tests

AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
marcogagliardi: done

AdminWebsite/AdminWebsite/ClientApp/src/app/booking/summary/summary.component.ts
```diff
@@ -243,69 +240,82 @@ export class SummaryComponent implements OnInit, OnDestroy {
...
                     caseNumber: this.hearing.cases[0].number
                 });
+
                 const hearingDetailsResponse = await this.hearingService.saveHearing(this.hearing);
 
-                if (hearingDetailsResponse.status === BookingStatus.Failed.toString()) {
-                    this.hearing.hearing_id = hearingDetailsResponse.id;
-                    this.setError(`Failed to book new hearing for ${hearingDetailsResponse.created_by} `);
-                    return;
+                if (this.hasJudge(hearingDetailsResponse)) {
```
oliver-scott, line 246: if there's no judge then we still need to do the steps below (clone multi days, navigate to the booking confirmation page) but without polling for the status, which is specific to hearings with a judge.

probably best to refactor lines 252-259 into a separate method so they can also be called if there is no judge.

marcogagliardi: done

AdminWebsite/AdminWebsite/ClientApp/src/app/booking/summary/summary.component.ts
```diff
@@ -243,72 +240,96 @@ export class SummaryComponent implements OnInit, OnDestroy {
...
-                    return;
-                }
+                const hearingDetailsResponse = await this.hearingService.saveHearing(this.hearing);
 
-                if (this.hearing.multiDays) {
-                    this.logger.info(`${this.loggerPrefix} Hearing is multi-day`, {
-                        hearingId: hearingDetailsResponse.id,
-                        caseName: this.hearing.cases[0].name,
-                        caseNumber: this.hearing.cases[0].number
+                if (this.hasJudge(hearingDetailsResponse)) {
```
oliver-scott, line 246: just noticed - there is already a `judgeAssigned` variable in this class. Think we can re-use that here instead of creating a new function

AdminWebsite/AdminWebsite/ClientApp/src/app/booking/summary/summary.component.ts
```diff
@@ -243,72 +240,96 @@ export class SummaryComponent implements OnInit, OnDestroy {
...
-                        );
-                    } else {
-                        this.logger.info(`${this.loggerPrefix} Hearing has just one day, no remaining days to book`, {
-                            hearingId: hearingDetailsResponse.id,
-                            caseName: this.hearing.cases[0].name,
-                            caseNumber: this.hearing.cases[0].number
-                        });
-                    }
+                } else {
+                    await this.processMultiHearing(hearingDetailsResponse);
```
oliver-scott, line 256: could we add test coverage for when there's no judge?

AdminWebsite/AdminWebsite/ClientApp/src/app/booking/summary/summary.component.ts
marcogagliardi: this one is checking in hearing object instead my function is checking in the response object returning from the server. What should I check?

AdminWebsite/AdminWebsite/ClientApp/src/app/booking/summary/summary.component.ts
marcogagliardi: test added

AdminWebsite/AdminWebsite/ClientApp/src/app/booking/summary/summary.component.ts
marcogagliardi: done

oliver-scott (dismissed):

oliver-scott (approved):

PR #1001, opened 2022-11-09
VIH-9278 team leader filters non availability hours

will-craig:
Moved div outside of span (was supposed to be in last change but was undone when merging)


marcogagliardi (dismissed):
👌🏽

marcogagliardi (approved):
👌🏽

PR #1006, opened 2022-11-16
VIH-9462 turn on edit mode when add non availability hours is selected

will-craig:
https://tools.hmcts.net/jira/browse/VIH-9462

### Change description ###
Changed add date to also turn on edit mode. Also fixed an issue where you cant delete a slot that has just been generated; opens the prompt so that it can send a delete request to the server (does not need to send request to server as it is not created in the db yet) - so will just remove it without the prompt in this case.

```
[] Yes
[X] No
```


oliver-scott (approved):

marcogagliardi (approved):
👌🏽

PR #1015, opened 2022-11-30
VIH-9355 Fix issues creating non-availabilities

oliver-scott:
https://tools.hmcts.net/jira/browse/VIH-9355

### Change description ###
Issues fixed:
- After adding a new non-availability, then going to add a second one, the last 2 rows are highlighted for validation errors
- Can't delete a non-availability after adding it
- Clicking Cancel keeps the new non-availability in the table, even though it hasn't been added yet. Fix by switching to edit mode before we push the new row to nonWorkHours array

The fix for the first 2 issues is to reload the non availabilities from the database after creating a new one so that we can populate the id for the new record.

```
[ ] Yes
[x] No
```


marcogagliardi (approved):
👌🏽

PR #1054, opened 2023-01-05
VIH-9389 vho unallocated hearings component

will-craig:
https://tools.hmcts.net/jira/browse/VIH-9389

### Change description ###
Created new unallocated hearings component, to be shown to VHO on the admin-web dashboard

```
[ ] Yes
[X] No
```


AdminWebsite/AdminWebsite/ClientApp/src/app/dashboard/dashboard.component.html
```diff
@@ -1,3 +1,4 @@
+<app-unallocated-hearings></app-unallocated-hearings>
```
oliver-scott, line 1: Should this be behind the work allocation feature flag?

oliver-scott (commented):
All looks really good. Just one comment, not sure if it had been agreed already

AdminWebsite/AdminWebsite/ClientApp/src/app/dashboard/dashboard.component.html
will-craig: Ah yeah, it wont show up in production as you need the vho team lead access, which no one will have. But you're right I should probably add the feature flag to the component just incase..

oliver-scott (approved):
👍

PR #1059, opened 2023-01-12
VIH-9607 Get hearing conference status changes

muralikaturi:
Removed unused methods on user account service. Changes to the get hearing conference status method to rely on the created status of the booking before the call to video api to retrieve the telephone conference id.

**Before creating a pull request make sure that:**

- [ x] commit messages are meaningful and follow good commit message guidelines
- [ ] README and other documentation has been updated / added (if needed)
- [ x] tests have been updated / new tests has been added (if needed)

```
[ ] Yes
[x ] No
```


shaed-parkar (approved):
We should try to align the testing libraries we use so they're more consistent, otherwise this looks solid

PR #1076, opened 2023-01-25
VIH-9366 vhotl navigates to and searches for hearings

will-craig:
### Change description ###
was using name instead of form-control-name and was being missed from request parameter

```
[ ] Yes
[X] No
```


shaed-parkar (approved):

shaed-parkar:
Might be worth adding a test to catch that or amending any existing one if it's giving a false positive

PR #1106, opened 2023-02-02
VIH-9679 capture all exceptions in middleware

shaed-parkar:

VIH-9679

### Change description ###

fix exception middleware
fix some code smells
update api clients

```
[ ] Yes
[x] No
```


AdminWebsite/AdminWebsite/ClientApp/src/tslint.json
```diff
@@ -1,5 +1,8 @@
 {
   "extends": "../tslint.json",
+  "linterOptions": {
+    "exclude": ["**/api-client.ts"]
```
oliver-scott, line 4: any reason we're doing this? We don't have this in place for the other apps

oliver-scott (commented):
just one comment, not sure if it was intended or not

AdminWebsite/AdminWebsite/ClientApp/src/tslint.json
shaed-parkar: something has changed recently with nswag and so the tslint disable and tslint disable are no longer the first lint and so lining keeps failing. I checked against a few versions and it keeps doing it. It's a file we don't really care to lint because it's autogenerated anyway

oliver-scott (approved):

PR #1109, opened 2023-02-07
VIH-9621 Store user profile in property instead of session

oliver-scott:
https://tools.hmcts.net/jira/browse/VIH-9621

### Change description ###

- Like Video Web, store the cached user profile in a property rather than the session to prevent users from tampering with their roles
- Add spec.ts for the logout component

```
[ ] Yes
[x] No
```


shaed-parkar (approved):
I really like this idea

PR #1112, opened 2023-02-09
VIH-9401 launch darkly keyvault change

will-craig:
Updated chart to get launch darkly client id from infra keyvalut rather than app one


marcogagliardi (approved):
👌🏽

PR #1120, opened 2023-02-20
VIH-9708 404 errors on GetTelephoneConferenceIdById endpoint from Admin web api

will-craig:
https://tools.hmcts.net/jira/browse/VIH-9708

### Change description ###
Added optional parameter to GetConferenceDetailsByHearingID, to allow querying of closed hearings. This is to cater for the edge case that a vho is still on the hearing details page of a hearing, just after it closes and they request the telephone conference ID which will return a 404 if the hearing has closed. Should not matter whether or not the hearing has closed in this particular request, as it's only fetching database data (telephone no.)

```
[ ] Yes
[X] No
```


shaed-parkar (approved):
LGTM

PR #1132, opened 2023-03-02
VIH-9738 - Previous-Message-Cleared-When-New-Search-Starts

RosDiaf:
https://tools.hmcts.net/jira/browse/VIH-9738

### Change description ###
The message 'Hearings have been updated.' needs to be cleared when:

- Allocated to user is selected
- Hearing is selected
- The same hearings is reallocated the message is cleared and auto refreshed when the other user is confirmed.


AdminWebsite/AdminWebsite/ClientApp/src/app/work-allocation/allocate-hearings/allocate-hearings.component.ts
```diff
@@ -43,6 +43,7 @@ export class AllocateHearingsComponent implements OnInit {
     customIconTransform: Transform = { rotate: 45 };
     private filterSize = 20;
     dropDownUserLabelAllocateTo = 'Allocate to';
+    readonly HEARING_HAVE_BEEN_UPDATED = 'Hearings have been updated.';
```
oliver-scott, line 46: we have a Constants file (in src/app/common/constants.ts) which we use for messages. Could we store it in there instead?

with an entry something like:

`AllocateHearings: {
        Confirmed: 'Hearings have been updated.'
    }`
    
but good idea nonetheless though, moving it into a constant :)

AdminWebsite/AdminWebsite/ClientApp/src/app/work-allocation/allocate-hearings/allocate-hearings.component.spec.ts
```diff
@@ -454,4 +454,47 @@ describe('AllocateHearingsComponent', () => {
             expect(clockIcon).toBeFalsy();
         });
     });
+
+    describe('Clear Hearing Updated Message', () => {
+        let HEARING_HAVE_BEEN_UPDATED: string;
+        let OTHER_MESSAGES: string;
+        beforeAll(() => {
```
oliver-scott, line 461: `beforeAll()` shouldn't be needed here, think you can change the `let`s to

```
const HEARING_HAVE_BEEN_UPDATED: string = 'Hearings have been updated.';
const OTHER_MESSAGES: string = 'Other messages';
```

and then remove the `beforeAll`

oliver-scott (commented):
looks good, added a couple of suggestions

AdminWebsite/AdminWebsite/ClientApp/src/app/work-allocation/allocate-hearings/allocate-hearings.component.ts
RosDiaf: Yes, I was looking for such a file.
Confimation message moved to constants file now.

AdminWebsite/AdminWebsite/ClientApp/src/app/work-allocation/allocate-hearings/allocate-hearings.component.spec.ts
RosDiaf: const added and beforeAll() removed

marcogagliardi (approved):
👌🏽

oliver-scott (approved):

PR #1136, opened 2023-03-03
VIH-9306 validate UK phone numbers for justice users

shaed-parkar:

VIH-9306

### Change description ###

- use same regex pattern for justice form name as add participant
- use a new NewUserRequest dto in admin web since the form has fewer fields than what Bookings API needs and reduce the coupling between Bookings API and Admin Web

```
[ ] Yes
[x] No
```

### DIFF
--- a/AdminWebsite/AdminWebsite.AcceptanceTests/packages.lock.json
+++ b/AdminWebsite/AdminWebsite.AcceptanceTests/packages.lock.json
@@ -152,8 +152,8 @@
       },
       "BookingsApi.Client": {
         "type": "Transitive",
-        "resolved": "1.43.13",
-        "contentHash": "tV08pKNr8ddSXxZQN0/GU91MQS2polyopMXv9o7yw2KOgvoF8XcIBX/kgmNQ9VCU2uQepWxU10SQo7tLeNi3Bg==",
+        "resolved": "1.43.16-pullrequest0611-0004",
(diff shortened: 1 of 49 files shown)


twyfordr273 (changes requested):
1. The telephone field should probably be updated to "tel" type rather than the "number" type. Number type would not allow brackets '(', ')', which should be valid in a telephone number. We have the pattern included to allow the brackets, so we should probably have the field match to allow those characters.

2. For the fields that are required we should probably add a required validator. We could also add a bit of css to indicate certain fields are required (A red asterisk or something similar)

twyfordr273:
> 2. For the fields that are required we should probably add a required validator. We could also add a bit of css to indicate certain fields are required (A red asterisk or something similar)

Based on GDS guidelines, the asterisk will not be added. So we can forget about this one.

twyfordr273 (dismissed):

twyfordr273:
There are two small code smells that could be gotten rid of. But everything else is good

komaldabhi1509 (approved):

oliver-scott (dismissed):

marcogagliardi (dismissed):
👌🏽

marcogagliardi:
only 2 code smell but not a blocker. good job

komaldabhi1509 (dismissed):

oliver-scott (approved):

will-craig (approved):

PR #1154, opened 2023-03-21
VIH-9789 Recording in CACD case type should always be set to off

arellison:
### **JIRA link**
https://tools.hmcts.net/jira/browse/VIH-9789

### **Change description**
Court of Appeal Criminal Division hearing should always have recordings turned off, even if an interpreter is booked on. To resolve this the audio_recording_required (boolean) now has an additional condition that is only set  to true if an interpreter is present and the case type does not equal "Court of Appeal Criminal Division.


AdminWebsite/AdminWebsite/ClientApp/src/app/booking/summary/summary.component.ts
```diff
@@ -93,7 +93,8 @@ export class SummaryComponent implements OnInit, OnDestroy {
         this.retrieveHearingSummary();
         this.switchOffRecording = this.recordingGuardService.switchOffRecording(this.hearing.case_type);
         this.interpreterPresent = this.recordingGuardService.mandatoryRecordingForHearingRole(this.hearing.participants);
-        this.hearing.audio_recording_required = this.interpreterPresent ? true : this.hearing.audio_recording_required;
+        this.hearing.audio_recording_required =
+            this.interpreterPresent && this.caseType !== 'Court of Appeal Criminal Division' ? true : this.hearing.audio_recording_required;
```
marcogagliardi, line 97: do we have only the caseType name or we have an ID for the case type? I am concern about adding a hard coded string instead of using an ID.

arellison: @marcogagliardi Hi Marco, at this point we only have caseType as a string. I looked to see if an ID was available, but there doesn’t appear to be. CaseType is used as a string in other parts of the code, such as RecordingGuardService - ClientApp\src\app\services\recording-guard.service.ts Let me know your thoughts, Andrew

arellison:
https://vh-admin-web-pr-1154.dev.platform.hmcts.net/

AdminWebsite/AdminWebsite/ClientApp/src/app/booking/summary/summary.component.ts
```diff
@@ -113,10 +113,20 @@ export class SummaryComponent implements OnInit, OnDestroy {
         this.bookinConfirmed = this.hearing.status === 'Created';
     }
 
+    setAudioRecordingRequired(currentValue: boolean): boolean {
```
oliver-scott, line 116: as this function purely returns a value, could we maybe name it `isAudioRecordingRequired` instead?

AdminWebsite/AdminWebsite/ClientApp/src/app/booking/summary/summary.component.ts
```diff
@@ -113,10 +113,20 @@ export class SummaryComponent implements OnInit, OnDestroy {
...
+        // CACD hearings should always have recordings set to off
+        if (this.caseType === 'Court of Appeal Criminal Division') {
```
oliver-scott, line 118: just to aid future re-use, could we maybe move this case type to `constants.ts`?

something like:

`CaseTypes: { CourtOfAppealCriminalDivision: 'Court of Appeal Criminal Division' }`

we could also reference this from the associated test as well.

oliver-scott (commented):
looks good, just a few suggestions

AdminWebsite/AdminWebsite/ClientApp/src/app/booking/summary/summary.component.ts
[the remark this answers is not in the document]
arellison: Thank you for your feedback. I have now made this change.

arellison: I have now made this change.

AdminWebsite/AdminWebsite/ClientApp/src/app/booking/summary/summary.component.ts
arellison: I have now made this change and added the reference constants.ts

oliver-scott (commented):
looks good, think you'll just need to run `npm run lint-fix` to fix the prettier issues

oliver-scott (approved):

marcogagliardi (approved):
👌🏽

phildebrah (approved):
Tested and working fine

PR #1201, opened 2023-05-23
VIH-9918 fix for toPromise codesmell 

will-craig:
updated all uses of deprecated toPromise to lastvaluefrom as recommended by rxjs docs


shaed-parkar (approved):

phildebrah (approved):
Tested and worked successfully

PR #1204, opened 2023-05-24
VIH-9924 form validation for existing user

shaed-parkar:

VIH-9924

### Change description ###

Somehow an existing user's phone number was invalid and when used in a future booking, the form's add participant button did not appear despite the phone number being invalid. The only way to display the error was to focus on on the field and leave it. Now the form is marked as dirty/touched when pre-populating to ensure the existing values are still considered valid.

```
[ ] Yes
[x] No
```


oliver-scott (approved):

phildebrah (approved):
Worked and Approved

PR #1208, opened 2023-05-31
Update cache to an empty array if a justice user does not exist

shaed-parkar:

VIH-9933

### Change description ###

* Update cache to an empty array if a justice user does not exist
* Move away from midddleware to OnTokenValidated Event to avoid multiple async calls to Bookings API

```
[ ] Yes
[x] No
```


phildebrah (approved):
seen

will-craig (approved):
lg2m

PR #1224, opened 2023-06-29
VIH-9970 trailing white spaces from existing persons fail a booking

shaed-parkar:

VIH-9970

### Change description ###

trim string values when re-populating add participant form from existing user

```
[ ] Yes
[#] No
```


oliver-scott:
just wondering about this - we already have some code that does this for creating a booking: https://github.com/hmcts/vh-bookings-api/blob/master/BookingsApi/BookingsApi/Controllers/HearingsController.cs#L773

could we leverage that instead, for consistency? Might need to move it to a common place, or mirror it

trimming as part of the API will protect us against requests made from external systems (like S & L)

shaed-parkar:
So that sanity check came in later and didn't protect the app from existing persons. I've run through this manually now so the trimming in admin-web as done by this pr gets past the booking step and the API correctly cleans the input and updates the entry.

oliver-scott (approved):

PR #1226, opened 2023-07-03
VIH-9775 - update nuget test package

marcogagliardi:
[VIH-9775](https://tools.hmcts.net/jira/browse/VIH-9775)


phildebrah (approved):
Tested and Approved

PR #1251, opened 2023-08-18
VIH-10087 Add new case type for Chester Crime Pathfinder

oliver-scott:
https://tools.hmcts.net/jira/browse/VIH-10087

### Change description ###
Disables recording for the Chester Crime case type

```
[ ] Yes
[x] No
```


will-craig (approved):

AdminWebsite/AdminWebsite/ClientApp/src/app/services/recording-guard.service.ts
```diff
@@ -5,7 +5,7 @@ import { ParticipantModel } from '../common/model/participant.model';
     providedIn: 'root'
 })
 export class RecordingGuardService {
-    excludedCaseTypes: string[] = ['Court of Appeal Criminal Division'];
+    excludedCaseTypes: string[] = ['Court of Appeal Criminal Division', 'Crime Crown Court'];
```
shaed-parkar, line 8: To avoid hard coding things like this, we should add a property to the case type to state whether or not it should be recorded, similar to the properties for hearing venues work allocation enabled and is Scottish

oliver-scott: I agree. Probably best to cover in a separate ticket

phildebrah (approved):
Tested and approve

PR #1249, opened 2023-08-10
VIH-10054 Handle editing hearings with expired case or hearing types

shaed-parkar:
### JIRA link (if applicable) ###

VIH-10054

### Change description ###

Handle editing hearings with expired case or hearing types

**Does this PR introduce a breaking change?** (check one with "x")

```
[ ] Yes
[x] No
```

PR #1254, opened 2023-08-22
VIH-0000 enable searching by email prior to valid email

shaed-parkar:

N/A

### Change description ###

enable searching by email prior to valid email being entered because partial match is allowed

```
[ ] Yes
[x] No
```


shaed-parkar:
fixed in #1249

PR #1263, opened 2023-09-06
VIH-10152 Audio recording required being changed to true on edit for Crime Crown Court and CACD

oliver-scott:
https://tools.hmcts.net/jira/browse/VIH-10152

### Change description ###
Fixes an issue where the Audio Recording Required flag is changed from false to true when editing a booking with an interpreter for the Crime Crown Court or CACD case types. It should always be set to false for these case types.

The logic for whether this flag should be true or false lives in the frontend. However the backend duplicates this and overrides the value in the request if there are interpreters. This is not needed.

```
[ ] Yes
[x] No
```


marcogagliardi (approved):
👌🏽

komaldabhi1509 (approved):

PR #1268, opened 2023-09-13
VIH-10158 Error adding panel member to booking

oliver-scott:
https://tools.hmcts.net/jira/browse/VIH-10158

### Change description ###
Fixes an issue adding a panel member when ejud is toggled on.
- Check for nulls before trimming whitespace

```
[ ] Yes
[x] No
```


AdminWebsite/AdminWebsite/ClientApp/src/app/booking/add-participant-base/add-participant-base.component.ts
```diff
@@ -216,10 +216,10 @@ export abstract class AddParticipantBaseDirective extends BookingBaseComponent i
             party: this.participantDetails.case_role_name,
             role: this.participantDetails.hearing_role_name,
             title: this.participantDetails.title ?? this.constants.PleaseSelect,
-            firstName: this.participantDetails.first_name.trim(),
-            lastName: this.participantDetails.last_name.trim(),
-            email: this.participantDetails.email.trim() || '',
-            phone: this.participantDetails.phone.trim() || '',
+            firstName: this.participantDetails.first_name?.trim(),
```
shaed-parkar, line 219: Since these fields are not editable when selecting an existing person from the search field, I can't help but think we should be doing something more than optional chaining.

shaed-parkar (commented):
Unsure if we should allow empty fields here. Rather filter search results or allow the field to be editable

shaed-parkar (approved):

PR #1270, opened 2023-09-15
VIH-10160 Added new booking status to display in booking list and detail views

will-craig:
https://github.com/hmcts/vh-bookings-api/pull/721

### Change description ###
Utilised the Ejud no judge badge, for the new booking statuses

```
[ ] Yes
[X] No
```


shaed-parkar (approved):
Looks good but the comment might be worth looking at if it makes sense and you have the time

AdminWebsite/AdminWebsite/ClientApp/src/app/bookings-list/booking-status/booking-status.component.ts
```diff
@@ -0,0 +1,28 @@
+import { Component, Input } from '@angular/core';
+import { BookingsDetailsModel } from '../../common/model/bookings-list.model';
+@Component({
+    selector: 'app-booking-status',
+    templateUrl: './booking-status.component.html',
+    styleUrls: ['./booking-status.component.css']
+})
+export class BookingStatusComponent {
+    @Input() bookingDetails: BookingsDetailsModel;
```
shaed-parkar, line 9: In the chance the whole model changes, this input would change too. It would be safer to have the input be status of type HearingStatus. That way you could even re-use this component in the booking details page and pass in the status irrespective of the model for that page

AdminWebsite/AdminWebsite/ClientApp/src/app/common/model/bookings-list.model.ts
```diff
@@ -105,4 +105,12 @@ export class BookingsDetailsModel {
     public get hasBookingConfirmationFailed(): boolean {
         return this.Status === 'Failed';
     }
+
+    public get hasBookingWithNoJudge(): boolean {
```
shaed-parkar, line 109: Do any components still use this?

AdminWebsite/AdminWebsite/ClientApp/src/styles.css
```diff
@@ -143,18 +142,12 @@
     float: left;
 }
 
+.vh-created-booking {
+    background-color: #3868c2;
```
shaed-parkar, line 146: use `govuk-colour("<colour name>")` instead. different versions have slightly changed the colours. If you use the function you simplify maintenance by not having to check the colour

shaed-parkar (commented):
Looks good, couple of points worth considering

AdminWebsite/AdminWebsite/ClientApp/src/app/bookings-list/booking-status/booking-status.component.ts
will-craig: the booking detail page and booking list page status' are a little differnt. The detail page isn't supposed to show the 'no judge' badge. We're also using the booking detail model proprties to select the css elements. Also the start time and duration from the model are displayed in this component view (because the sit conveniently between the NO JUDGE badge and the status text

AdminWebsite/AdminWebsite/ClientApp/src/app/common/model/bookings-list.model.ts
will-craig: it's being used in AdminWebsite/AdminWebsite/ClientApp/src/app/bookings-list/booking-status/booking-status.component.html for selecting the css class. I debated whether this method would best be place with that component. But I decided this is logic for the the model itself, and should go with the other status methods.

AdminWebsite/AdminWebsite/ClientApp/src/styles.css
will-craig: I used the hex code to avoid changing the file type to scss, and keep the component dependency minimal, but I actually I suppose it doent matter too much. Ill update it

AdminWebsite/AdminWebsite/ClientApp/src/app/bookings-list/booking-status/booking-status.component.ts
will-craig: Deleted my previous comment, you're right I'm not actually using it anywhere anymore. Ill remove it

AdminWebsite/AdminWebsite/ClientApp/src/app/common/model/bookings-list.model.ts
will-craig: Deleted my previous comment, you're right I'm not actually using it anywhere anymore. Ill remove it

AdminWebsite/AdminWebsite/ClientApp/src/app/bookings-list/booking-status/booking-status.component.ts
will-craig: We're using other properties from the BookingDetailsModel, such as the time and duration and the flags for the status? Are you saying we should make a model for the input? The status property on the BookingDetailsModel is just a string. Also this status component is slightly differnt to the one shown in the booking detail view, as that one doesnt show the no judge badge.

komaldabhi1509 (approved):

PR #1316, opened 2023-12-11
VIH-10363 Fix judge name disappearing on booking list when V2 switched on

oliver-scott:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/VIH-10363

### Change description ###
2 changes needed:

1. Get the judge name from judiciary participants if the judge is booked as a judiciary participant
2. Update `is_judge` and `is_staff_member` to look at the user role name. The case and hearing role are null when V2 is switched on so instead read from user role name, which is populated for both V1 and V2


marcogagliardi (approved):
👌🏽

komaldabhi1509 (approved):

PR #1347, opened 2024-02-05
VIH-10469 MRD casetypes not showing in dropdown

will-craig:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/VIH-10469

### Change description ###
Updated the controller that returns the case types to also include case types without hearing types. Once the V2 feature toggle is turned on will likely need to revisit this with better design, once the hearing type entity is completely redundant.

### DIFF
--- a/AdminWebsite/AdminWebsite.UnitTests/Controllers/ReferenceDataControllerTests.cs
+++ b/AdminWebsite/AdminWebsite.UnitTests/Controllers/ReferenceDataControllerTests.cs
@@ -142,7 +142,7 @@ namespace AdminWebsite.UnitTests.Controllers
             }
         }
 
-        private List<CaseTypeResponse> GetCaseTypesList()
+        private static List<CaseTypeResponse> GetCaseTypesList()
         {
             return new List<CaseTypeResponse>
             {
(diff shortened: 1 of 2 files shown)


AdminWebsite/AdminWebsite.UnitTests/Controllers/ReferenceDataControllerTests.cs
```diff
@@ -143,42 +146,48 @@ private List<CaseTypeResponse> GetCaseTypesList()
...
+                new()
+                {
+                    Id = 4, Name = "type4", ServiceId = "AA4",
+                    HearingTypes = new List<HearingTypeResponse>()
                 }
             };
         }
 
-        private List<HearingTypeResponse> GetHearingTypes()
+        private List<AdminWebsite.Contracts.Responses.HearingTypeResponse> GetHearingTypes()
```
oliver-scott, line 182: could we make this method static to remove the sonar cloud warning?

oliver-scott (commented):
just one minor suggestion, otherwise looks good :)

AdminWebsite/AdminWebsite.UnitTests/Controllers/ReferenceDataControllerTests.cs
will-craig: Yeah, sure thing

oliver-scott (approved):

komaldabhi1509 (approved):

PR #1349, opened 2024-02-08
VIH-10349 Updated nuget package references

will-craig:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/VIH-10349

### Change description ###
Updated updateStatus endpoint and angular components to reflect changes to booking api update endpoint.


AdminWebsite/AdminWebsite/Controllers/HearingsController.cs
```diff
@@ -870,12 +867,12 @@ public async Task<IActionResult> UpdateBookingStatus(Guid hearingId, UpdateBooki
         [ProducesResponseType(typeof(UpdateBookingStatusResponse), (int) HttpStatusCode.OK)]
         [ProducesResponseType((int) HttpStatusCode.NotFound)]
         [ProducesResponseType((int) HttpStatusCode.BadRequest)]
-        public async Task<IActionResult> UpdateHearingStatus(Guid hearingId)
+        public async Task<IActionResult> FailBooking(Guid hearingId)
```
oliver-scott, line 870: where is this called from? I can't see it in the angular code

will-craig: It's called from the video-hearing.service.ts line 163 "updateFailedStatus()" -- this hasnt actually changed at all, in which is why it's not in the commit. All I've done is update the controller method name, as it's only used for this one purpose (setting the booking status to failed) rather than setting any status

oliver-scott: ah I see, sorry yes I was looking for failBooking

oliver-scott (approved):

komaldabhi1509 (approved):

PR #1362, opened 2024-02-28
VIH-10513 Removed logic for v1 endpoint to pull eJudiciary account information resulting in generic accounts that exists in both tables showing

will-craig:


AdminWebsite/AdminWebsite/Controllers/JudiciaryAccountsController.cs
```diff
@@ -53,31 +53,14 @@ public async Task<ActionResult<IList<JudgeResponse>>> PostJudgesBySearchTermAsyn
             {
                 term = _encoder.Encode(term);
                 var searchTerm = new SearchTermRequest(term);
-
                 List<JudgeResponse> allJudges;
                 var courtRoomJudgesTask = _userAccountService.SearchJudgesByEmail(searchTerm.Term);
+                var courtRoomJudges = await courtRoomJudgesTask;
```
oliver-scott, line 58: could we remove `_featureToggles` now that it is no longer used?

oliver-scott (commented):
looks good to me, just one suggestion

oliver-scott (approved):

komaldabhi1509 (approved):

PR #1367, opened 2024-03-11
VIH-10530 Fix - Error for enhanced edit of multi-day booking with no participants

oliver-scott:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/VIH-10530

### Change description ###
The `MapUpdateJudiciaryParticipantsRequestV2` method skips judiciary participants that have not changed in order to avoid an additional call to bookings API when updating hearings. However we don't want to do this for the enhanced update of multi-day bookings, as this is already handled on the bookings API side and results in a bad request from not sending any participants.

Refactor the mapping of existing judiciary participants into a separate method


marcogagliardi (approved):
lg2m

komaldabhi1509 (approved):

PR #1373, opened 2024-04-05
VIH-10553 Conditional updates for multi hearing edits

oliver-scott:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/VIH-10553

### Change description ###
The multi-day hearing edit feature allows you to edit one day of the multi-day, and then apply those updates to all future days in the multi-day group.

This PR modifies the admin web backend to only apply updates that are different between the existing hearing and the contents of the request. These partial updates can then be applied to the future days of the multi-day hearing rather than overwriting them with what is in the request, since the future days may have different details, participants etc to the hearing being edited if they have been edited individually.

Additionally, V1 support has been dropped for this feature, so we only show the update button in the UI when using the V2 API.


azure-pipelines.sds.pr-release.yml
```diff
@@ -5,7 +5,7 @@ resources:
     - repository: azTemplates
       type: github
       name: hmcts/azure-devops-templates
-      ref: master
+      ref: feature_sa_fix
```
oliver-scott, line 8: Temporary fix provided by DevOps for the Docker Compose issue following an update from Microsoft. Can be reverted once Microsoft have rolled the update back

will-craig (approved):

komaldabhi1509 (approved):

PR #1385, opened 2024-05-08
VIH-10674 Fix freeze clicking edit button when judiciary participants have same display name

oliver-scott:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/VIH-10674

### Change description ###

- Sort by email after display name to ensure the two sorting mechanisms sort the judicial members in the same order. Fixes an issue where the UI infinitely attempts to re-sort them and causing the UI to freeze
- Update nuget packages, update code for removed user api method (`GetJudges`)
- Revert changes to helm chart (values.dev.template.yaml)


AdminWebsite/AdminWebsite/ClientApp/src/app/booking/participant/list/participant-list.component.ts
```diff
@@ -37,18 +37,22 @@ export class ParticipantListComponent implements OnInit, OnChanges, DoCheck {
...
         const judiciaryEmailListChanged = JSON.stringify(judicialMembersLocal) !== JSON.stringify(sortedJudicialMembersLocal);
         if (judiciaryEmailListChanged) {
-            this.sortJudiciaryMembers();
+            this.sortJoh();
         }
     }
 
+    sortJoh() {
+        this.sortJudiciaryMembers();
+    }
```
will-craig, line 54: Why are wrapping sortJudiciaryMembers() in another function, rather than calling directly?

will-craig (commented):
Looks good, just one comment from me.

AdminWebsite/AdminWebsite/ClientApp/src/app/booking/participant/list/participant-list.component.ts
oliver-scott: it's purely for the unit tests, so that we can do this:

```
spyOn(component, 'sortJoh').and.callFake(() => {
                sortJudiciaryMembersCalled = true;
                component.sortJudiciaryMembers();
            });
```

I wanted to mock `sortJudiciaryMembers()` so that we can set sortJudiciaryMembersCalled = true when it is called. However we still need to invoke the actual code within `sortJudiciaryMembers()` so that the judiciary members are still sorted.

as you can't spy on a method and have it call itself, I had to separate it into 2 methods, so that we can spy on `sortJoh()` while having it call the actual sorting method (`sortJudiciaryMembers()`).

will-craig (approved):

komaldabhi1509 (approved):

PR #1413, opened 2024-07-12
VIH-10816 Fix for error updating representative

oliver-scott:
### Jira link (if applicable)
https://tools.hmcts.net/jira/browse/VIH-10816

### Change description ###

- Remove V1 logic from `IsRoleRepresentative` (which uses case role/party). This doesn't work on V2
- Update tests to use V2 roles
- Fix flush error


shaed-parkar (approved):
Looks simple enough. Nice to work through some house keeping, even if it's through bug tickets 😅

komaldabhi1509 (approved):

PR #1421, opened 2024-08-29
VIH-10799 Updated nuget packages and removed redundant unused polly retry method from conference service

will-craig:


AdminWebsite/AdminWebsite/Services/ConferenceDetailsService.cs
```diff
@@ -1,62 +1,24 @@
...
-        private readonly IVideoApiClient _videoApiClient;
-
-        public ConferenceDetailsService(IPollyRetryService pollyRetryService, ILogger<ConferenceDetailsService> logger, IVideoApiClient videoApiClient)
-        {
-            _pollyRetryService = pollyRetryService;
-            _logger = logger;
-            _videoApiClient = videoApiClient;
-        }
-
-        public async Task<ConferenceDetailsResponse> GetConferenceDetailsByHearingIdWithRetry(Guid hearingId, string errorMessage)
```
shaed-parkar, line 34: Isn't this implementation used by the front end after a booking to poll for a conference?

shaed-parkar (approved):

AdminWebsite/AdminWebsite/Services/ConferenceDetailsService.cs
will-craig: It used to be when it polled within the admin-web bff, the functionality was moved within the angular client itself and it polls from there instead, and makes individual requests each time. This was left over and isn't used anywhere anymore.

PR #1458, opened 2024-12-20
VIH-11019 Consolidate ng models for bookings

oliver-scott:
### Jira link
https://tools.hmcts.net/jira/browse/VIH-11019

### Change description
**Key changes**
- Remove booking details model, participant model and participant details model. Replace with VHBooking and VHParticipant
- Move the api contract to model mappers into separate class (api-contract-to-client-model-mappers.ts)


AdminWebsite/AdminWebsite/ClientApp/src/app/common/model/vh-participant.ts
```diff
@@ -0,0 +1,132 @@
...
+    isInterpretee?: boolean;
+    userRoleName?: string;
+    isCourtroomAccount?: boolean;
+    addedDuringHearing?: boolean;
+    contactEmail?: string;
+    isJudiciaryMember?: boolean;
+    interpretation_language: InterpreterSelectedDto;
+    screening?: ScreeningDto;
+    // flag to indicate if participant is the last in the list and don't need decoration bottom line
+    flag?: boolean;
```
shaed-parkar, line 36: angular ngFor loops have this built-in using the `isLast` field. If it's just for display purposes better to leave it with ng instead of us managing it
```
<div *ngFor="let item of items; let isLast=last">
  {{ item }}
  <ng-container *ngIf="isLast">This is the last item!</ng-container>
</div>
```

oliver-scott: thanks, this was from the original view model that was migrated across. There might be other properties like this that were just added for convenience

oliver-scott: have removed this property

AdminWebsite/AdminWebsite/ClientApp/src/app/booking/add-participant-base/add-participant-base.component.ts
```diff
@@ -198,9 +198,9 @@ export abstract class AddParticipantBaseDirective extends BookingBaseComponent i
         this.displayErrorNoParticipants = false;
         this.displayAdd();
         this.enableFields();
-        this.participantDetails = { ...participantDetails };
+        this.participantDetails = participantDetails.clone();
```
shaed-parkar, line 201: Why use a clone method? Could the object spread operator such as:
```suggestion
        this.participantDetails = {...participantDetails};
```

AdminWebsite/AdminWebsite/ClientApp/src/app/common/model/vh-participant.ts
```diff
@@ -0,0 +1,126 @@
...
+    screening?: ScreeningDto;
+
+    constructor(init?: Partial<VHParticipant>) {
+        Object.assign(this, init);
+        if (!this.externalReferenceId) {
+            this.externalReferenceId = uuid();
+        }
+    }
+
+    get fullName(): string {
```
shaed-parkar, line 43: I wouldn't typically suggest xml docs on fields but it might be worth clearing up that full name include Title

AdminWebsite/AdminWebsite/ClientApp/src/app/testing/stubs/search-email-stub.ts
```diff
@@ -1,10 +1,10 @@
 import { Component, Output, ViewChild, ElementRef, EventEmitter } from '@angular/core';
-import { ParticipantModel } from '../../common/model/participant.model';
+import { VHParticipant } from 'src/app/common/model/vh-participant';
```
shaed-parkar, line 2: I've found ngMocks useful in testing that has dependencies on other components such as this. If it's an easy removal, then it would be nice to see the back of this

AdminWebsite/AdminWebsite/ClientApp/src/app/booking/add-participant-base/add-participant-base.component.ts
oliver-scott: I had this originally, but unfortunately the spread operator doesn't clone the getters

AdminWebsite/AdminWebsite/ClientApp/src/app/common/model/vh-participant.ts
oliver-scott: added

AdminWebsite/AdminWebsite/ClientApp/src/app/testing/stubs/search-email-stub.ts
oliver-scott: good idea, could I suggest a separate ticket for that though. There are quite a few instances of stubs which would be to good to replace en masse rather than making this PR bigger

shaed-parkar (approved):

PR #1466, opened 2025-01-16
VIH-10156 Replace hardcoded case types for audio recording exclusion

oliver-scott:
### Jira link
https://tools.hmcts.net/jira/browse/VIH-10156

### Change description

- `RecordingGuardService` - replace the hardcoded list of excluded case types with the IsAudioRecordingAllowed property
- Change the hearing api response to return the case type properties as part of a case type object
- Add some additional mappers and tests


AdminWebsite/AdminWebsite.IntegrationTests/packages.lock.json
```diff
@@ -104,18 +104,18 @@
       },
       "BookingsApi.Client": {
         "type": "Transitive",
-        "resolved": "3.1.14",
-        "contentHash": "j+zkb9f2uRlGWS3Jte2ES43XfBO/r434V8WmcDYm5wBoy5acl26f1bPLbd0mXH/bjd78BDhAu6Oo7LXl5cN2JQ==",
+        "resolved": "3.1.17-pr-0939-0005",
```
oliver-scott, line 107: TODO update after bookings api merge

charts/vh-admin-web/values.dev.template.yaml
```diff
@@ -8,4 +8,5 @@ java:
     AZUREAD__POSTLOGOUTREDIRECTURI: https://${SERVICE_FQDN}/logout
     AZUREAD__REDIRECTURI: https://${SERVICE_FQDN}/home
     DOM1__POSTLOGOUTREDIRECTURI: https://${SERVICE_FQDN}/logout
-    DOM1__REDIRECTURI: https://${SERVICE_FQDN}/home
\ No newline at end of file
+    DOM1__REDIRECTURI: https://${SERVICE_FQDN}/home
+    VHSERVICES__BOOKINGSAPIURL: https://vh-bookings-api-pr-939.dev.platform.hmcts.net/
```
oliver-scott, line 12: TODO revert before merge

AdminWebsite/AdminWebsite/Mappers/HearingTypeResponseMapper.cs
```diff
@@ -0,0 +1,16 @@
...
+using BookingsApi.Contract.V2.Responses;
+
+namespace AdminWebsite.Mappers;
+
+public static class HearingTypeResponseMapper
+{
+    public static HearingTypeResponse Map(this CaseTypeResponseV2 response) =>
+        new()
+        {
+            Group = response.Name,
```
shaed-parkar, line 11: Do we still need to use group? Feels like hang over from V1. How much would it be to rename group to name so the prop names line up and are up to date in terms of what they represent?

shaed-parkar (approved):
Looks good!

AdminWebsite/AdminWebsite/Mappers/HearingTypeResponseMapper.cs
oliver-scott: I've created a separate ticket for this. There's still a bit of terminology hangover: https://tools.hmcts.net/jira/browse/VIH-11250

ronaldmalone (approved):

PR #1473, opened 2025-02-07
VIH-10558 user api clean up

will-craig:
### Jira link
https://tools.hmcts.net/jira/browse/VIH-10558

### DIFF
--- a/charts/vh-admin-web/values.dev.template.yaml
+++ b/charts/vh-admin-web/values.dev.template.yaml
@@ -9,6 +9,4 @@ java:
     AZUREAD__REDIRECTURI: https://${SERVICE_FQDN}/home
     DOM1__POSTLOGOUTREDIRECTURI: https://${SERVICE_FQDN}/logout
     DOM1__REDIRECTURI: https://${SERVICE_FQDN}/home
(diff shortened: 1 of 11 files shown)


shaed-parkar (approved):

charts/vh-admin-web/values.dev.template.yaml
```diff
@@ -8,4 +8,5 @@ java:
     AZUREAD__POSTLOGOUTREDIRECTURI: https://${SERVICE_FQDN}/logout
     AZUREAD__REDIRECTURI: https://${SERVICE_FQDN}/home
     DOM1__POSTLOGOUTREDIRECTURI: https://${SERVICE_FQDN}/logout
-    DOM1__REDIRECTURI: https://${SERVICE_FQDN}/home
\ No newline at end of file
+    DOM1__REDIRECTURI: https://${SERVICE_FQDN}/home
+
```
shaed-parkar, line 12: reset this to master if there's no real change

shaed-parkar (approved):
Can't believe some of this code still lived

charts/vh-admin-web/values.dev.template.yaml
will-craig: I used this pr to point at the user-api, have removed the template url which is why it's showing a change

